PR #7, opened 2019-12-23
AvroDecoding :: Add support for nullable List of nested records

brsimoes:
Hi,

The issue you've solved for encoding nested null records needs also to be addressed when decoding (https://github.com/sksamuel/avro4k/issues/2)

I'm submitting a pull request to address it. Can you check it please?

Thanks in advance.

Best regards,
Bruno Diaw


sksamuel:
Sorry I don't know how I missed this PR!.

Issue #8, opened 2019-12-26
Documentation needed: serializing Timestamps et. al.


psarka:
I can't figure out how to use the types mentioned in that big "Types" table. For example, simple code

```
import com.sksamuel.avro4k.Avro
import kotlinx.serialization.Serializable
import java.sql.Timestamp


@Serializable
data class Example(
    val time: Timestamp
)

fun main() {
    println(Avro.default.schema(Example.serializer()).toString())
}
```

fails with an error
```
Serializer has not been found for type 'Timestamp'. To use context serializer as fallback, explicitly annotate type or property with @ContextualSerialization
```

If you can tell me how to make it work, I'll open a PR with doc improvement :)

PR #16, opened 2020-01-17
Update README.md with LogicalType annotation explanation

psarka:
I'm not sure if that's the intended usage, but that's the best I could come up with (since no one bothered to reply to #8)


sksamuel:
Thank you

PR #22, opened 2020-02-10
Add @AvroDefault annotation

alexhwoods:
Hey @sksamuel! Thank you for your work on this project; it's a great library.

For this feature, I followed the API provided by the [`@AvroDefault`](https://avro.apache.org/docs/1.8.0/api/java/org/apache/avro/reflect/AvroDefault.html) annotation. 

I was a little hesitant to have the function take in a `String`, but it is one of the simpler choices, and there's already an existing API that does it, so I think it's fine. I believe they define it as a JSON string.

Ideally it'd be a generic type, but you can't use generic types with annotation classes.

# How It Works
* Added the `@AvroDefault` in the same format as the other annotations in the project
* the key lines are these:
```kotlin
val default: Any? = annos.default()?.let {
    when(fieldDescriptor.kind) {
        PrimitiveKind.INT -> it.toInt()
        PrimitiveKind.LONG -> it.toLong()
        PrimitiveKind.FLOAT -> it.toFloat()
        PrimitiveKind.BOOLEAN -> it.toBoolean()
        PrimitiveKind.BYTE -> it.toByte()
        PrimitiveKind.SHORT -> it.toShort()
        PrimitiveKind.STRING -> it
        PrimitiveKind.UNIT -> null
        else -> throw IllegalArgumentException("Cannot use a default value for type ${fieldDescriptor.kind}")
    }
}

val field = Schema.Field(fieldNaming.name(), schemaWithResolvedNamespace, annos.doc(), default)
```
* added tests

### DIFF
--- a/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
+++ b/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
@@ -19,5 +19,6 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
    fun doc(): String? = annotations.filterIsInstance<AvroDoc>().firstOrNull()?.value
    fun aliases(): List<String> = annotations.filterIsInstance<AvroAlias>().map { it.value }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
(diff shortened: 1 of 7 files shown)


arosenb2:
Totally need this for a Kafka Connect app I’m working on!

sksamuel:
Great, love it.

PR #23, opened 2020-02-13
Migrate to kotlinx.serialization 0.14

chriswk:
Started using this library together with ktor's ContentNegotiation with kotlinx.serialization. Unfortunately 0.20 couldn't be used with ktor 1.3.0 as ktor brings along kotlinx.serialization:0.14, causing the comparison for UnionKind.SEALED to fail, because the UnionKind.SEALED no longer exists in 0.14.

This PR bumps kotlinx.serialization to 0.14, and makes the necessary override changes for List/Record Encoder and Decoders.


sksamuel:
Awesome thanks.

Issue #25, opened 2020-02-24
AvroAlias is not repeatable


thake:
Hey @sksamuel, first of all, thanks for the great library! I really like the easy way of serialization.

I'm currently writing a maven plugin that does compile avro schema files to avro4k compatible kotlin code as requested by #19 . 

I tried to support record aliases but it seems that the Defintion of "@AvroAlias" does not support repetitive annotations. As multiple aliases are a feature of avro, I think this is a bug in avro4k.

PR #26, opened 2020-02-24
Added new annotation @AvroAliases to support multiple aliases.

thake:
Fix for #25 

Adding a new AvroAliases annotation that supports multiple aliases as in avro spec. I didn't reuse the AvroAlias annotation for this, because a list of strings is closer to the Avro spec.


sksamuel:
Nice work

Issue #27, opened 2020-02-25
AvroDefault does not work for a avro null default


thake:
Consider the following data class definition:

```kotlin
@Serializable
data class BarString(
   val a: String,
   @AvroDefault("hello")
   val b: String,
   @AvroDefault("null")
   val nullableString : String?,
   @AvroDefault("hello")
   val c:String?
)
```

the correct schema would be according to avro specs:
```javascript
{
   "type": "record",
   "name": "BarString",
   "namespace": "com.sksamuel.avro4k.schema",
   "fields": [
      {
         "name": "a",
         "type": "string"
      },
      {
         "name": "b",
         "type": "string",
         "default": "hello"
      },
      {
         "name": "nullableString",
         "type": ["null","string"],
         "default" : null
      },
      {
         "name": "c",
         "type": ["string","null"],
         "default": "hello"
      }
   ]
}

```
Please take a look at the order within the union of the "nullableString" field. The avro null type is first. This is because default values of union types can only be an instance of the first type within the union.  

Avro4k currently fails when generating the schema:

> Invalid default for field nullableString: "null" not a ["null","string"]
org.apache.avro.AvroTypeException: Invalid default for field nullableString: "null" not a ["null","string"]
	at org.apache.avro.Schema.validateDefault(Schema.java:1540)
	at org.apache.avro.Schema.access$500(Schema.java:87)
	at org.apache.avro.Schema$Field.<init>(Schema.java:521)
	at org.apache.avro.Schema$Field.<init>(Schema.java:557)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.buildField(ClassSchemaFor.kt:99)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.dataClassSchema(ClassSchemaFor.kt:36)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.schema(ClassSchemaFor.kt:24)
	at com.sksamuel.avro4k.Avro.schema(Avro.kt:231)
	at com.sksamuel.avro4k.schema.AvroDefaultSchemaTest$1.invokeSuspend(AvroDefaultSchemaTest.kt:13)

sksamuel:
0.22.0 released.

PR #28, opened 2020-02-25
Fixed incorrect handling of avro null default values.

thake:
If a default is specified for a union, the default must be an instance of the first type in the union. Thus the generated schema needs to check the default definition when generating the union schema.

Additionally added a Avro.NULL constant that should be used to indicate a avro null default.

This is a fix for #27


sksamuel:
Could the annotation just accept a null to indicate a null default?

thake:
Unfortunately null values are not allowed in annotations.

Stephen Samuel <notifications@github.com> schrieb am Mi., 26. Feb. 2020,
14:39:

sksamuel:
:+1:

PR #40, opened 2020-03-23
Use avro4k-core package name in README

ivoanjo:
Hello there 👋

Just a small fix for something that had me confused for a few minutes :)


sksamuel:
Ty

PR #54, opened 2020-04-24
Provide default value for type 'array'

### LINKED ISSUE #52: Default value for lists
Hi

Upon using @AvroDefault annotation, I noticed that there is no option to set a default empty list value.

```kotlin
when (fieldDescriptor.kind) {
               PrimitiveKind.INT -> it.toInt()
               PrimitiveKind.LONG -> it.toLong()
               PrimitiveKind.FLOAT -> it.toFloat()
               PrimitiveKind.BOOLEAN -> it.toBoolean()
               PrimitiveKind.BYTE -> it.toByte()
               PrimitiveKind.SHORT -> it.toShort()
               PrimitiveKind.STRING -> it
               else -> throw IllegalArgumentException("Cannot use a default value for type ${fieldDescriptor.kind}")
            }
```

Could StructureKind.LIST be supported so that the default value could be an empty [] array.

See  [Avro default value spec](https://avro.apache.org/docs/current/spec.html)

Thank you

daicode-jan:
Provide:
- null default for arrays
- empty list default for arrays
- primitive default for arrays

example:
```kotlin
@Serializable
data class BarArray(
   @AvroDefault(Avro.EMPTY_LIST)
   val defaultEmptyArray: List<String>,
   @AvroDefault(Avro.NULL)
   val nullableDefaultEmptyArray: List<String>?,
   @AvroDefault("1,2")
   val defaultStringArrayWith2Defaults: List<String>,
   @AvroDefault("1,2")
   val defaultIntArrayWith2Defaults: List<Int>,
   @AvroDefault("3.14, 9.89")
   val defaultFloatArrayWith2Defaults: List<Float>
)
```

Fixes #52

### DIFF
--- a/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
+++ b/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
@@ -25,6 +25,5 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
       }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
    fun default(): String? = annotations.filterIsInstance<AvroDefault>().firstOrNull()?.value
-   fun defaultArray(): Array<String>? = annotations.filterIsInstance<AvroDefaultList>().firstOrNull()?.value
 }
 
--- a/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
+++ b/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/AnnotationExtractor.kt
@@ -25,5 +25,6 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
       }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
    fun default(): String? = annotations.filterIsInstance<AvroDefault>().firstOrNull()?.value
+   fun defaultArray(): Array<String>? = annotations.filterIsInstance<AvroDefaultList>().firstOrNull()?.value
 }
 
--- a/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/Avro.kt
+++ b/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/Avro.kt
@@ -95,7 +95,6 @@ class Avro(override val context: SerialModule = EmptyModule) : SerialFormat, Bin
        * Use this constant if you want to explicitly set a default value of a field to avro null
        */
       const val NULL = "com.sksamuel.avro4k.Avro.AVRO_NULL_DEFAULT"
-      const val EMPTY_LIST = "com.sksamuel.avro4k.Avro.AVRO_EMPTY_LIST_DEFAULT"
    }
 
    /**
--- a/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/Avro.kt
+++ b/avro4k-core/src/main/kotlin/com/sksamuel/avro4k/Avro.kt
@@ -90,11 +90,11 @@ class Avro(override val context: SerialModule = EmptyModule) : SerialFormat, Bin
          UUID::class to UUIDSerializer())
       )
       val default = Avro(simpleModule)
+
       /**
        * Use this constant if you want to explicitly set a default value of a field to avro null
        */
       const val NULL = "com.sksamuel.avro4k.Avro.AVRO_NULL_DEFAULT"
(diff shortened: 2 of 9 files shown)


daicode-jan:
Ha yes, good catch. I pushed a second commit to povide  an alternative annotation that gets rid of the simplistic .split

```kotlin
@Serializable
data class BarArray(
   @AvroDefaultList([])
   val defaultEmptyArray: List<String>,
   @AvroDefault(Avro.NULL)
   val nullableDefaultEmptyArray: List<String>?,
   @AvroDefaultList(["John", "Doe"])
   val defaultStringArrayWith2Defaults: List<String>,
   @AvroDefaultList(["1", "2"])
   val defaultIntArrayWith2Defaults: List<Int>,
   @AvroDefaultList(["3.14", "9.89"])
   val defaultFloatArrayWith2Defaults: List<Float>
)
```

- @AvroDefault(Avro.NULL) still works on lists

thake:
Thanks for your contribution! The patch looks really good and seems like a
good fit for default values of arrays. However considering additional cases
such as default values of records, I can't see a way to have a consistent
api if we go with an array specific annotation.

How about using the default string as it is used in an avro schema. This
would be easy to parse for lists and is later extendable for other complex
types such as maps and records. See
https://avro.apache.org/docs/current/spec.html#schema record

For lists this would lead to a annotation like this:
@AvroDefault("""
["a","b"]
""")

What do you think about this?



daicode-jan <notifications@github.com> schrieb am Fr., 24. Apr. 2020, 21:52:

daicode-jan:
In order to keep using the @AvroDefault(string) I propose treating the defaultValue as json for arrays.

```kotlin
@Serializable
data class BarArray(
   @AvroDefault(Avro.EMPTY)
   val defaultEmptyArray: List<String>,
   @AvroDefault(Avro.NULL)
   val nullableDefaultEmptyArray: List<String>?,
   @AvroDefault("""["John", "Doe"]""")
   val defaultStringArrayWith2Defaults: List<String>,
   @AvroDefault("""["1", "2"]""")
   val defaultIntArrayWith2Defaults: List<Int>,
   @AvroDefault("""["3.14", "9.89"]""")
   val defaultFloatArrayWith2Defaults: List<Float>
)

   private fun transformStringToTypedArray(serialDescriptor: SerialDescriptor, stringArray: String): List<Any> {
      val value = try {
         objectMapper.readTree(stringArray.toByteArray())
      } catch (ioe: IOException) {
         throw IllegalArgumentException("Cannot use set default value $stringArray, invalid array")
      }
      if (!value.isArray) throw IllegalArgumentException("Default value '$stringArray' for type ${serialDescriptor.kind} does not result into an array.")
      val array = objectMapper.treeToValue(value, Any::class.java) as ArrayList<*>
      return array.map { toArrayOfType(serialDescriptor, "$it") }
   }
````

We could either map that array to the specified type and be able to throw an exception if the user made a mistake in the json or blindly return whatever came out of the jsonNode

**Comment by thake** [comment]
Sounds like a good idea. The implementation for maps and records as default values should be easy using the json syntax. 

One note though. You should leverage the JSON serialization of kotlinx.serialization instead of jackson. Doing so would automatically reuse all custom serializers (for records). Also it does not add another dependency.

**Comment by daicode-jan** [comment]
Doing the parsing with kotlinx.serialization requires adding @ImplicitReflectionSerializer to 3 classes.

```kotlin
   private fun transformStringToTypedArray(fieldDescriptor: SerialDescriptor, stringArray: String): List<Any> = try {
      json.toJson(stringArray).jsonArray.map { it.content }
   } catch (ioe: IOException) {
      throw IllegalArgumentException("Cannot use default value $stringArray for type ${fieldDescriptor.kind}, invalid json array")
   }
```

using Json with default config. Is there something that I am missing?

sksamuel:
Doesn't avro come with jackson anyway?

daicode-jan:
it does

sksamuel:
Might as well just use that then to keep it simple ?

thake:
But you'll still have to make sure that your data class can correctly be
deserialized using kotlinx.serialization and Jackson if you want to use it
in a default value. I'll look into a way to use kotlinx serialization
without reflection tomorrow.

Stephen Samuel <notifications@github.com> schrieb am So., 26. Apr. 2020,
22:02:

daicode-jan:
@thake I managed to do it, I was not initially familiar with kotlinx.serialization

Updated PR

thake:
That looks good. I just have a few minor modifications. Can you open your pull request for modifications?

daicode-jan:
Not familiar with github but "Allow edits from maintainers" is checked

thake:
I added a few tests for Array and Set types. Also, I removed the constant for an empty list default value as this can easily be represented as String. The Avro.NULL default value only exists because annotations cannot contain null values and there is no other way to express a null value.
Does this work for you?

PR #75, opened 2020-10-19
Reworked Avro encoding and decoding API

thake:
Reworked Avro encoding and decoding API in order to have type safety for required encoding / decoding settings.

The following code would perfectly compile:
```kotlin
val input = openInputStream<T>(serializer) {
    format = AvroFormat.BinaryFormat
}
```

but at runtime an exception will be thrown because the readerSchema needs to be set for BinaryFormat.

To prevent this wrong API usage, the API has been redesigned so that the compiler can detect these errors. The call would look like this with the new API:
```kotlin
val input = openInputStream<T>(serializer) {
   decodeFormat = AvroDecodeFormat.Binary(mySchema)
}


thake:
@sksamuel I changed the API a little bit and removed the deprecated bits. As you designed the API, I would be very happy if you can take a look at the changes.

thake:
Hey @sksamuel, do you think you'll have the time to look at this PR? Otherwise, I will just go forward and merge this PR soon.

sksamuel:
If you're happy to merge it, then please do.
I'm back from vacation now, and just catching up. I will setup the creds
tomorrow or Tuesday and add github actions to release.

PR #90, opened 2020-12-28
Fix #89 support default values for enum types

### LINKED ISSUE #89: Default Value for Enum
Hi, I would like to use the default feature for enums that is described here : [https://avro.apache.org/docs/current/spec.html#Enums](https://avro.apache.org/docs/current/spec.html#Enums).

I tried to use the [AvroDefault](https://github.com/avro-kotlin/avro4k/blob/master/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt#L60) annotation, giving it the String value of an enum value, but I get a NumberFormatException when I try that.
I'd like to use this avro feature so that schemas that use enums can be backward compatible. It seems the only way to achieve this is to give the field a default value that is one of the enum values in case a reader is using an older version that does not know about added values for that enum.

I'm pretty new to avro4k, so maybe avro4k supports this feature in a way I don't know about.

Here is what I tried:

```kotlin
@Serializable
enum class IngredientType { VEGGIE, MEAT, }

@Serializable
data class Ingredient(val name: String, @AvroDefault("MEAT") val type: IngredientType,
    val sugar: Double, val fat: Double,)

@Serializable
data class Pizza(val name: String, val ingredients: List<Ingredient>, val vegetarian: Boolean,
    val kcals: Int,)

fun main() {
  val schema = Avro.default.schema(Pizza.serializer())
  println(schema.toString(true))
}
```
This is what I get:

```
Exception in thread "main" java.lang.NumberFormatException: Character M is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:518)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:401)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:834)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.convertToAvroDefault(ClassSchemaFor.kt:135)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.buildField(ClassSchemaFor.kt:114)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.dataClassSchema(ClassSchemaFor.kt:55)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.schema(ClassSchemaFor.kt:43)
	at com.sksamuel.avro4k.schema.ListSchemaFor.schema(SchemaFor.kt:88)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.buildField(ClassSchemaFor.kt:76)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.dataClassSchema(ClassSchemaFor.kt:55)
	at com.sksamuel.avro4k.schema.ClassSchemaFor.schema(ClassSchemaFor.kt:43)
	at com.sksamuel.avro4k.Avro.schema(Avro.kt:239)
	at co.da.avro.DomainKt.main(Domain.kt:18)
	at co.da.avro.DomainKt.main(Domain.kt)
```

If this feature is not supported in avro4k, do you plan to support it in the future?

kossi:
Fixes this issue that avro4k did not handle default values in enums https://github.com/avro-kotlin/avro4k/issues/89

@thake btw should avro4k use something like https://ktlint.github.io/ to enforce the formatting and some rules. As it seems it is now a bit mixed set of lines that follows the editorconfig file that the project has and probably some are IDE/editor defaults.


thake:
@kossi First of all thanks for the PR! I looked into the issue and your patch actually seems to mix field defaults and enum default (a fallback default value for unknown symbols, see #91). 
To fix #89 it should be enough to add Schema.Type.ENUM to the line 121 in ClassSchemaFor. To fix #91 there should be a new PR.

kossi:
> @kossi First of all thanks for the PR! I looked into the issue and your patch actually seems to mix field defaults and enum default (a fallback default value for unknown symbols, see #91).
> To fix #89 it should be enough to add Schema.Type.ENUM to the line 121 in ClassSchemaFor. To fix #91 there should be a new PR.

No it doesn't? If you add default there then the schema would look something like this -> 

```json
{
  "type" : "record",
  "name" : "EnumWithDefaultTest",
  "namespace" : "com.github.avrokotlin.avro4k.schema.EnumSchemaTest",
  "fields" : [ {
    "name" : "type",
    "type" : {
      "type" : "enum",
      "name" : "IngredientType",
      "namespace" : "com.github.avrokotlin.avro4k.schema",
      "symbols" : [ "VEGGIE", "MEAT" ]
    },
    "default" : "MEAT"
  } ]
}
``` 

That is not correct as the spec defines that defaults for enum should be on the same level as symbols are. There was same issue with the original project avro4s 

https://github.com/sksamuel/avro4s/issues/413

kossi:
Hi @thake, can you look this again.

Or I think that solution would not have been efficient if there is no way to pass default for the actual enumeration type. So  yes changes the ClassSchemaFor just to use the normal annotation case. And also added a separate annotation "AvroEnumDefault" with what you can get default for the schema that at least the current spec documentation seems to support.

@darmstrong1 also

src/main/kotlin/com/github/avrokotlin/avro4k/schema/SchemaFor.kt
```diff
@@ -37,14 +38,33 @@ interface SchemaFor {
...
+   private val descriptor: SerialDescriptor
+) : SchemaFor {
    override fun schema(): Schema {
       val naming = RecordNaming(descriptor)
       val entityAnnotations = AnnotationExtractor(descriptor.annotations)
       val symbols = (0 until descriptor.elementsCount).map { descriptor.getElementName(it) }
-      val enumSchema = SchemaBuilder.enumeration(naming.name).doc(entityAnnotations.doc()).namespace(naming.namespace).symbols(*symbols.toTypedArray())
+
+      val defaultSymbol = entityAnnotations.enumDefault()?.let { fieldDefault ->
+         if (fieldDefault != Avro.NULL) {
```
thake, line 52: Avro.NULL is no valid default for an enum default, therefore the value within the annotation should always match a enum symbol.

src/test/kotlin/com/github/avrokotlin/avro4k/schema/EnumSchemaTest.kt
```diff
@@ -48,4 +97,16 @@ enum class Wine {
...
-}
\ No newline at end of file
+}
+
+@Serializable
+@AvroEnumDefault("MEAT")
+enum class IngredientType { VEGGIE, MEAT, }
+
+@Serializable
+@AvroEnumDefault("MEAT")
```
thake, line 107: Annotating a data class with an AvroEnumDefault should have no impact on the gernerated schema for the data class, as the AvroEnumDefault annotation should only be evaluated for enums. Is this what you want to test with this case? If so, please clarify it in the test name.

src/main/kotlin/com/github/avrokotlin/avro4k/schema/ClassSchemaFor.kt
```diff
@@ -118,6 +119,7 @@ class ClassSchemaFor(
       val default: Any? = annos.default()?.let {
          when {
              it == Avro.NULL -> Schema.Field.NULL_DEFAULT_VALUE
+             schemaWithResolvedNamespace.extractNonNull().type == Schema.Type.ENUM -> annos.default()
```
thake, line 122: You can just add it to the list in the following line as "it" is annos.default()

thake (changes requested):
Thanks for the fast update :+1: 

Your solution looks very good, I just have some minor suggestions. After merging this PR #89 and #91 can be closed. Great!

src/main/kotlin/com/github/avrokotlin/avro4k/schema/SchemaFor.kt
kossi: Thanks removed. Or forgot the check the actual implementation or this was earlier checking those field defaults that isn't true anymore. And changed the local variable to enumDefault to be more correct

src/main/kotlin/com/github/avrokotlin/avro4k/schema/ClassSchemaFor.kt
kossi: Moved this check to use same list check

src/test/kotlin/com/github/avrokotlin/avro4k/schema/EnumSchemaTest.kt
kossi: Actually I just removed this test. Or it was something that @darmstrong1 added to test field defaults. But with the new enum defaults annotation and the original avrodefault annotation probably it sufficient enough that you don't need to anymore add that additional layer / field?

thake (approved):

thake:
I just added another test to check that the error described in #89 does not occur.

The code looks good :+1: . Thank you very much!

darmstrong1:
@kossi,

Thank you for doing this!

PR #93, opened 2021-01-04
Upgrade kotlin 1.4.10 -> 1.4.21 

kossi:
Updated kotlin to latest and avro minor version bump. 

Also there were some compiler warnings that bothered me. Probably nothing major but added compiler argument for that kotlin OptIn and changed Arvo class to use file suppression to fix import warning. Or now the build is free from warnings and allWarningsAsErrors kotlin option could be used.

I wonder why that dependency-bot doesn't automatically upgrade these?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,7 @@ plugins {
    id("org.jetbrains.kotlin.plugin.serialization") version Libs.kotlinVersion
    id("maven-publish")
    signing
-   id("org.jetbrains.dokka") version Libs.kotlinVersion
+   id("org.jetbrains.dokka") version Libs.dokkaVersion
    id("io.kotest") version Libs.kotestGradlePlugin
 }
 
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -51,6 +51,7 @@ dependencies {
 tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    kotlinOptions.jvmTarget = "1.8"
```
kossi, line 52: Btw is there some reason that the jvm target is still 1.8 and not 11? And especially the apiVersion of kotlin? Or I don't what is the policy of libraries usually or the compatibility

thake: Java 1.8 as no feature of Java 9+ is used. As a lot of programs are still written / maintained in Java 1.8 this seems a reasonable choice as a library.
But you are right with the kotlin api version. I think it should be changed to 1.4 as the kotlinx.serialization 1.x won't be working with a lower kotlin api.

thake:
:+1:

Issue #72, opened 2020-10-13
Support for snake-casing


williamboxhall:
It looks like Avro4s has this https://github.com/sksamuel/avro4s#field-mapping

It would help us a lot if we could use it in Avro4k.

thake:
Sounds like a good feature! Do you think that you could provide a PR for this?

thake:
Implemented in release v1.1.0

PR #94, opened 2021-01-07
Support for different casing (snake, pascal)

kossi:
Hellou,

Added support for snake and pascal casing, there is issue related to this #72 .  I'm not 100% sure if this is going to handle every case or how this should be tested.

Probably some people doesn't generate avro schemas from kotlin classes or example my personal use is at the moment a bit different. Anyways this could be quite useful if we can get avro4k to support different casing. To bad kotlin doesn't have implicit imports like scala etc or I was bit confused from where the end-user should actually pass the casing but @thake perhaps you have better knowledge how this should be implemented.

Sorry about the network related examples and please enforce the editorconfig in the build or something. Or I'm it scared to touch some of the files as my editor. And it seems that the current .editorconfig and some personal editor setups are quite different.


sksamuel:
@thake this looks good to me.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -86,14 +87,17 @@ open class AvroInputStreamBuilder<T>(
 class AvroDeserializerInputStreamBuilder<T>(
    private val deserializer: DeserializationStrategy<T>,
    private val avro : Avro,
+   private val namingStrategy: NamingStrategy,
    converter: (Any) -> T
 ) : AvroInputStreamBuilder<T>(converter) {
-   val defaultReadSchema : Schema by lazy { avro.schema(deserializer.descriptor) }
+   val defaultReadSchema : Schema by lazy { avro.schema(deserializer.descriptor, namingStrategy) }
```
thake, line 93: I actually wouldn't want the user to pass a NamingStrategy to the AvroDeserializerInputBuilder / AvroOutputStreamBuilder). 
How about having the namingStrategy as a property of the Avro instance? It would be an implementation detail of the Avro class that shouldn't be known to classes like AvroDeserializerInputBuilder.

src/main/kotlin/com/github/avrokotlin/avro4k/decoder/ListDecoder.kt
```diff
@@ -15,9 +16,11 @@ import org.apache.avro.generic.GenericArray
...
 
 @ExperimentalSerializationApi
-class ListDecoder(private val schema: Schema,
-                  private val array: List<Any?>,
-                  override val serializersModule: SerializersModule
+class ListDecoder(
+   private val schema: Schema,
+   private val array: List<Any?>,
+   override val serializersModule: SerializersModule,
+   private val namingStrategy: NamingStrategy,
```
thake, line 23: How about wrapping the namingStrategy in an (not yet existing) "AvroConfiguration" object? It would make it possible to easily introduce new configuration options and propagate them to all Decoder / Encoder.

src/main/kotlin/com/github/avrokotlin/avro4k/schema/SchemaFor.kt
```diff
@@ -39,7 +39,7 @@ class EnumSchemaFor(
    private val descriptor: SerialDescriptor
 ) : SchemaFor {
    override fun schema(): Schema {
-      val naming = RecordNaming(descriptor)
+      val naming = RecordNaming(descriptor, DefaultNamingStrategy)
```
thake, line 42: Why the DefaultNamingStrategy?

src/main/kotlin/com/github/avrokotlin/avro4k/decoder/SealedClassDecoder.kt
```diff
@@ -27,8 +30,8 @@ class SealedClassDecoder (descriptor: SerialDescriptor,
    private var currentState = DecoderState.BEFORE
 
    var leafDescriptor : SerialDescriptor = descriptor.leavesOfSealedClasses().firstOrNull {
-      val schemaName = RecordNaming(value.schema.fullName, emptyList())
-      val serialName = RecordNaming(it)
+      val schemaName = RecordNaming(value.schema.fullName, emptyList(), DefaultNamingStrategy)
```
thake, line 33: Maybe it would be good to introduce a new invoke method on RecordNaming using the DefaultNamingStrategy. Just from reading this code one might get the impression that passing "DefaultNamingStrategy" at this point is incorrect because namingStrategy exists. But as the value.schema.fullName has already been decoded using the namingStrategy, passing the DefaultNamingStrategy is actually correct. Having a new invoke method without a naming strategy would imply that the name will not be changed.

src/main/kotlin/com/github/avrokotlin/avro4k/schema/ClassSchemaFor.kt
```diff
@@ -28,7 +28,7 @@ class ClassSchemaFor(
 ) : SchemaFor {
 
    private val entityAnnotations = AnnotationExtractor(descriptor.annotations)
-   private val naming = RecordNaming(descriptor)
+   private val naming = RecordNaming(descriptor, DefaultNamingStrategy)
```
thake, line 31: Wouldn't the namingStrategy be needed here? Otherwise, the name of the Schema wouldn't be changed according to the strategy.

thake (changes requested):
@kossi I'm very sorry for such a late review. I just started a new contract this year and did not find the time. 
Thanks again for your great pull requests! As part of this PR I would like to introduce an AvroConfiguration class that holds all serialization relevant configuration options. One of these options is the NamingStrategy. Having a wrapper around the currently only option gives us the possibility to add further options without significant refactoring. The new AvroConfiguration should be an optional parameter to the "Avro" class.

What do you think about this?

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
kossi: Yes that probably makes sense or I wonder wondering where the user should pass this information

src/main/kotlin/com/github/avrokotlin/avro4k/decoder/SealedClassDecoder.kt
kossi: Added new invoke method. I wonder if it's a abit anti-pattern or should just default values be used. Or the whole RecordNaming class looks now a bit weird perhaps?

src/main/kotlin/com/github/avrokotlin/avro4k/schema/SchemaFor.kt
kossi: I guess this actually should never call the naming strategy or this is the enum name. And probably you don't want to touch it? Or if change this  to use given strategy then example snake_casing will generate enum looking like this: 

``` 
{
  "name" : "v",
  "type" : {
    "type" : "enum",
    "name" : "_internet_protocol",
    "namespace" : "com.github.avrokotlin.avro4k.schema",
    "symbols" : [ "IPv4", "IPv6" ]
  }
}
```

Field level name v is okay but that name object probably should not be touched? Well I don't know how exactly python / etc components actually use those schemas to generates types.. Or does it use some kind of dict structure for avro objects dunno

src/main/kotlin/com/github/avrokotlin/avro4k/schema/ClassSchemaFor.kt
kossi: This also probably is a field that should not run through the naming strategy. Or same thing as with enums but now with the example it generates something like this for Interface record -> 

```
{ 
 "type" : "record",
  "name" : "_interface",
  "namespace" : "com.github.avrokotlin.avro4k.schema.NamingStrategySchemaTest",
 ...
```

kossi:
> @kossi I'm very sorry for such a late review. I just started a new contract this year and did not find the time.
> Thanks again for your great pull requests! As part of this PR I would like to introduce an AvroConfiguration class that holds all serialization relevant configuration options. One of these options is the NamingStrategy. Having a wrapper around the currently only option gives us the possibility to add further options without significant refactoring. The new AvroConfiguration should be an optional parameter to the "Avro" class.
> 
> What do you think about this?

No problem. Good luck with your new gig.

I think your suggestions sounds good or I implemented AvroConfiguration data class that is a field on the Avro class. And yes it makes those streamBuilders etc much nicer. Hopefully I understood it correctly.


Build are failing because of some dependencies. It is probably ok to bump latest? Is it ok also to upgrade gradle to 6.8?

I wonder should that dokka dependency be dropped for now as those docs are not actually generated. And when eventually somebody have time or knowledge to get them working then bring it back.

thake:
Thanks for the rework! I just made a small adjustment to the constructor.

PR #100, opened 2021-04-13
Trying to update kotest to latest version

thake:


thake:
@sksamuel can you have a look at this PR? I tried to update kotest to the latest available version but somehow the tests now run forever. Maybe you have a clue about what is going on as you are also maintaining kotest. 

Thanks in advance!

sksamuel:
Yep will take a look.

thake:
Fixed

PR #102, opened 2021-06-11
Migrate to Kotlin 1.5

vladimirfx:
Notes on PR:
- Kotlin 1.5.20-RC is used because 1.5.10 compiler crashed on tests compilcation
- mass changes in tests because Kotlinx Serialization dropped support for function-local classes


thake:
@vladimirfx Thanks for the PR! Looks very good :+1:

PR #106, opened 2021-07-14
If an optional field is missing from a previoius schema, do not blow up.

RyanAtWare2Go:
Avro introduced a change here https://issues.apache.org/jira/browse/AVRO-2278 that changes the way get works. This breaks backwards compattible schema evolution. Previously if you tried to get an optional field that did not exist on the schema, you just got null. Now it throws an exception


sksamuel:
lgtm

@thake ?

thake:
@RyanAtWare2Go Thanks for the great patch and sorry to let you wait!

PR #109, opened 2021-09-01
Polymorphic open support

williamboxhall:
Huge disclaimer: I don't know kotlinx or avro4k internals at all so there's a good chance this implementation is not on the right track. It was also done under delivery time pressure so forgive me for not doing deeper homework on this. Happy to take any pointers on how to improve it.

This PR adds support for open / unsealed / abstract classes, configured like so:
```kotlin
val module = SerializersModule {
   polymorphic(UnsealedPolymorphicRoot::class) {
      subclass(UnsealedChildOne::class)
      subclass(UnsealedChildTwo::class)
   }
}
val schema = Avro(serializersModule = module).schema(UnsealedPolymorphicRoot.serializer())
```

For now it explicitly doesn't support polymorphic defaults - this is a feature someone could build on top if they want to.

PR #110, opened 2021-09-06
Polymorphic open support (Take 2)

williamboxhall:
This is a replacement for https://github.com/avro-kotlin/avro4k/pull/109 - Forking this on my personal account to make it easier for @thake to make changes to it.

This PR adds support for open / unsealed / abstract classes to generate avro schemas and serialize/deserialize avro binary, configured like so:
```kotlin
val module = SerializersModule {
   polymorphic(UnsealedPolymorphicRoot::class) {
      subclass(UnsealedChildOne::class)
      subclass(UnsealedChildTwo::class)
   }
}
val schema = Avro(serializersModule = module).schema(UnsealedPolymorphicRoot.serializer())
```

For now it explicitly doesn't support polymorphic defaults - this is a feature someone could build on top if they want to.


williamboxhall:
Hey @thake, if it's still too hard to figure out collaboration on the [cultureamp branch PR](https://github.com/avro-kotlin/avro4k/pull/109) then we can just do it here instead - whatever works for you.

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/UnionEncoder.kt
```diff
@@ -16,7 +16,10 @@ import org.apache.avro.Schema
 class UnionEncoder(private val unionSchema : Schema,
                    override val serializersModule: SerializersModule,
                    private val callback: (Record) -> Unit) : AbstractEncoder() {
-
+   override fun encodeString(value: String){
+      //No need to encode the name of the concrete type. The name will never be encoded in the avro schema.
+      println(value)
```
williamboxhall, line 21: Did you mean to leave this `println` here @thake ?

src/main/kotlin/com/github/avrokotlin/avro4k/SerialDescriptor.kt
```diff
@@ -1,46 +1,25 @@
...
-      val collector = object : SerializersModuleCollector {
-         override fun <T : Any> contextual(kClass: KClass<T>, provider: (typeArgumentsSerializers: List<KSerializer<*>>) -> KSerializer<*>) {
-         }
-
-         override fun <Base : Any, Sub : Base> polymorphic(baseClass: KClass<Base>, actualClass: KClass<Sub>, actualSerializer: KSerializer<Sub>) {
-            captured.add(baseClass to actualSerializer)
-         }
+fun SerialDescriptor.possibleSerializationSubclasses(serializersModule: SerializersModule): List<SerialDescriptor> {
+    return when (this.kind) {
+        PolymorphicKind.SEALED -> this.leavesOfSealedClasses()
```
williamboxhall, line 20: I'm hoping without much work we should be able to get this working for sealed interfaces. I think the fix wouldn't be here but would be in the part of the code that detects the `kind` based on the input. Currently it's treating `sealed interface` as `PolymorphicKind.OPEN` rather than `PolymorphicKind.SEALED` but I'm not exactly sure where that detection is done - do you know how we might do that?

src/test/kotlin/com/github/avrokotlin/avro4k/schema/PolymorphicClassSchemaTest.kt
```diff
@@ -43,20 +38,6 @@ class PolymorphicClassSchemaTest : StringSpec({
...
-            subclass(UnsealedChildTwo::class)
-            default { TypeSerializer }
-         }
-      }
-      val exception = shouldThrow<SerializationException> {
-         Avro(serializersModule = module).schema(UnsealedPolymorphicRoot.serializer())
-      }
-      exception.message shouldBe "Polymorphic defaults are not supported in avro4k. Error whilst describing: UnsealedPolymorphicRoot"
-   }
-
```
williamboxhall: You don't think it's worth keeping this test kind of as documentation - in case people come looking trying to figure out if this is supported or not. Not strong feelings either way, though

williamboxhall (commented):
Looking great @thake - what do you see as the next steps towards getting this into a release?

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/UnionEncoder.kt
thake: Thanks, just needed for a breakpoint. :) I will remove it

src/main/kotlin/com/github/avrokotlin/avro4k/SerialDescriptor.kt
thake: I will give the support of sealed interfaces a try in a separate commit as my time resources are limited right now.

src/test/kotlin/com/github/avrokotlin/avro4k/schema/PolymorphicClassSchemaTest.kt
thake: I removed the exception if the SerializersModule contained a default subclass for polymorphic classes. Although default subclasses do not make sense for avro, I don't want to limit the reusability of SerializersModules across serialization formats. So I think the test does not make that much sense without the assertion for an exception.

thake:
@williamboxhall If you approve the changes made to your PR I will merge it and build a new release.

williamboxhall:
@thake I approve the changes and very much appreciate it!

thake:
@williamboxhall thank you very much for your work! I will prepare a relase.

williamboxhall:
Thanks @thake!

PR #117, opened 2021-09-25
Fix contextual serialization inside value classes

vladimirfx:
Fixes serialization of:

```Kotlin
@JvmInline
value class SomeValueClass(val a: TypeWithContextualSerializer)
```


thake:
Thanks for this fix @vladimirfx! :+1:

PR #126, opened 2022-02-22
Add support for using defaut values

asmadsen:
Add support for using defaut values in constructor if record does not contain field


Chuckame:
I think it should be configurable

thake:
@asmadsen, thanks for opening this PR!

While this change would improve usability with default values in kotlin, I think it conflicts with how Avro schemas should be read. Avro schemas already contain default values for optional fields. These default values are considered for encoding/decoding if the annotation `@AvroDefault` is used. 

The `@AvroDefault` annotation was needed because we cannot access the kotlin native default values. If we consider the kotlin native default value when decoding values, we would not be consistent with the generated Avro schema because the kotlin native default value is not part of the Avro read schema.

Look at `AvroDefaultValuesDecoderTest` for more information on how to use Avro default values.

thake:
Closed because of no active discussion

PR #132, opened 2022-10-07
fix: Allow value classes and primitive schemas generation

Chuckame:
I detected a bug while using a field `Map<MyStringValueClass, NestedObject>`. It is not handling well value classes everywhere. By the way I also added tests for primitive schema generation with also primitive schema that are coming from value classes.
It was because of the value class descriptor has not the STRING kind, while the elementDescriptor at index 0 is of STRING kind.


Chuckame:
Thanks @thake, BTW when should you deploy the new version including those changes?

thake:
I'll deploy it today. I'm currently working on making it possible to directly serialize primitives, enums, collections, and maps without having to wrap them in a record. @Chuckame, your contribution helped here a lot! Thanks!

Chuckame:
Ah ok, super! Do you need help to deserialize primitives ?

PR #142, opened 2023-05-25
implement encode/decode from/to single object format

jangalinski:
This implements  SingleObject as AvroEncodeFormat and AvroDecodeFormat. The In/Out Streams are realized using avros BinaryMessage[De|En]coder which reads/writes a GenericRecord to singleObjectBytes and uses the existing toRecord/fromRecord functions to convert to instance of data class <T>.


src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -22,34 +23,39 @@ import java.nio.file.Paths
...
 open class AvroInputStreamBuilder<T>(
    private val converter: (Any) -> T
-   ) {
+) {
    /**
     * The format that should be used to decode the bytes from the input stream.
     */
-   var decodeFormat : AvroDecodeFormat = defaultDecodeFormat
+   var decodeFormat: AvroDecodeFormat = defaultDecodeFormat
+
```
thake, line 31: Could you deactivate your auto-format? While I generally like a consistent code style, unfortunately, there is currently no consistent code style in this project. Thus applying auto format only makes it harder to read PRs because of the additional diffs.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -254,6 +274,51 @@ class Avro(
...
+    * Writes an instance of <T> using the given or derived writer [Schema].
+    * This method will use the [AvroEncodeFormat.SingleObject] format.
+    *
+    * @param serializer serialization strategy (as you get using `MyClass.serializer()`)
+    * @param writerSchema optional writer schema, if not given, it is derived from the serializer
+    * @param value the instance of type <T> we want to serialize.
+    *
+    * @return the written instance will be returned as a single object encoded [ByteArray] containing the schema fingerprint.
+    */
+   fun <T> encodeToSingleObject(
```
thake, line 287: I wouldn't add new decode and encode methods to the Avro class. Other encodings also do not have methods on this level. The only exception has been made for the "data" encoding. I'm thinking of deprecating the data methods to have a consistent API. So adding new methods would add new methods that need to be deprecated ^^.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -254,6 +274,51 @@ class Avro(
...
+    * using [AvroEncodeFormat.SingleObject]. The schema used will be the writer schema derived by the embedded fingerprint using [SchemaStore].
+    *
+    * @param deserializer deserialization strategy (as you get using `MyClass.serializer()`)
+    * @param readerSchema optional readerSchema, if not given, it is derived from the deserializer.
+    * @param schemaStore an avro [SchemaStore] that can lookup the writer schema used to encode the single object bytes.
+    * @param singleObjectEncodedBytes avro bytes in single object format
+    *
+    * @return instance of type <T> compliant with the reader schema containing values from binary payload.
+    */
+   fun <T> decodeFromSingleObject(
```
thake, line 313: see above

src/main/kotlin/com/github/avrokotlin/avro4k/io/AvroSingleObjectOutputStream.kt
```diff
@@ -0,0 +1,41 @@
...
+@OptIn(ExperimentalSerializationApi::class)
+class AvroSingleObjectOutputStream<T>(
+   private val output: OutputStream,
+   private val converter: (T) -> GenericRecord,
+   writerSchema: Schema
+) : AvroOutputStream<T> {
+
+   private val messageEncoder = BinaryMessageEncoder<GenericRecord>(GenericData.get(), writerSchema)
+
+   override fun flush() {
```
thake, line 28: call to `output.flush()` is missing.

thake (changes requested):
@jangalinski, thank you very much for this great addition. I just have some minor remarks, after that we are ready to merge.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
Chuckame: I'll add the necessary to unify the formatting rules automatically into another pr. Until that, as said, please commit only your changes

jangalinski: Will do. My autoformatting is set to respect the editorconfig settings, so this is a bit confusing to me ... but I will look into it.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
jangalinski: So for now, just using the in/out stream builders is the API? ok, can do.

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
Chuckame: Yes but it is kind of empty, we will clean it in another pr. For the moment, if you can only commit the real changes even if it's in another format to make less noise for the review, that would be great 🙏

jangalinski: done

jangalinski:
OK, I worked on all changes and refactored the test so it does encode/decode without using any specific Avro.default methods.

I kind of lost it during reformatting ... so this where more very small commits than I intended. Sorry for that.
But I assume you will squash anyway, right?

jangalinski:
BTW: We are starting to use and like avro4k ... but we have a lot of questions and ideas.

Would you be open to allowing discussions on this repo? I would like to avoid raising "Question: ..." issues en gros.

thake:
> BTW: We are starting to use and like avro4k ... but we have a lot of questions and ideas.
> 
> Would you be open to allowing discussions on this repo? I would like to avoid raising "Question: ..." issues en gros.

Thanks for the feedback :) I hope we can improve our documentation so that fewer questions stay open. 

I'll give the discussions section a try. I was hesitant to do so because now users have to choose whether `Issues` or `Discussions` is the right place for contact. But as I said, I'll give it a try ^^. Thanks for the suggestion.

thake (approved):

src/test/kotlin/com/github/avrokotlin/avro4k/io/AvroSingleObjectInputStreamTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+      val other: OtherFooString = Avro.default.openInputStream(OtherFooString.serializer()) {
+         decodeFormat = AvroDecodeFormat.SingleObject(
+            readerSchema = readerSchema,
+            schemaStore = schemaStore
+         )
+      }.from(ByteArrayInputStream(decodedBytes)).nextOrThrow()
+
+      // value is equal
+      other.str.shouldBeEqual(foo.str)
```
Chuckame, line 40: Sorry for the late review, I think it's better to test serialize+deserialize inside the same method. I think you can serialize with avro4k then deserialize with official avro library, and do the reverse (serialize with official avro library then deserialize with avro4k). That way you are sure that avro4k is compatible with the avro official library 😄

jangalinski: Hm. I do not really get your point. Do you mean The InputStreamTest and the OutputStreamTest should become one test class and use java->kotlinx and kotlinx->java de/serializations?

Wouldn't then a use-case based test that does not directly refer to one of the concrete stream implementations the better option? So on top level "SingleObjectSerializationTest" with everything related?

Or did I completly miss your intention?

jangalinski: Assuming that I got this right: I created a use case test: https://github.com/toolisticon/avro4k/blob/feature/141_single_object/src/test/kotlin/com/github/avrokotlin/avro4k/UseCaseSingleObjectEncodingTest.kt

This uses FooString (from the avro4k decoder-test) and FooStringJava (a java specificRecord I generated with the official maven plugin) to test all interoperability cases:

kotlin -> kotlin
kotlin -> java
java -> kotlin
java -> java

It is green ... let me know if this is what you had in mind.
If so, we can remove/refactor the additional tests for the In/Out streams.

It might be a good idea to also implement the other use cases (aka Decoding Formats) in this way.

Chuckame: Not really, I was talking about those tests:
- encoding data from pure java avro encoder (with the schema coming from Avro.schemaFor since we are not testing this here) then try to decode it using your new method in avro4k (GenericDatumWriter by example)
- and the reverse, encode data from your new method in avro4k (still using the same schema from avro4k) then decode it using pure java avro decoder (GenericDatumReader by example)
- In both tests, you want to check that the finally decoded data is the same as the previouslyencoded data

Why ? It prove that we are fully compliant with the pure java avro library. In your current tests, you are encoding and decoding the data only using your avro4k methods

src/test/kotlin/com/github/avrokotlin/avro4k/UseCaseSingleObjectEncodingTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+      kotlinDecoded shouldBeEqual fooStringKotlin
+   }
+
+   "encode fooStringJava and decode fooStringJava" {
+      val bytes = fooStringJava.toByteBuffer().array()
+      val javaDecoded = FooStringJava.fromByteBuffer(ByteBuffer.wrap(bytes))
+
+      javaDecoded shouldBeEqual fooStringJava
+   }
```
Chuckame, line 86: This test seems not really useful since it is not testing avro4k but the generated stuff

src/test/kotlin/com/github/avrokotlin/avro4k/UseCaseSingleObjectEncodingTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+      val kotlinDecoded = TestFixtures.decodeSingleObject(
+         bytes = bytes,
+         serializer = FooString.serializer(),
+         readerSchema = schemaFooStringKotlin,
+         schemaStore = schemaStore
+      )
+
+      kotlinDecoded shouldBeEqual fooStringKotlin
+   }
```
Chuckame, line 66: To be more clear: The tests you added feels great, but I was not talking about that.

The aim of avro4k is to encode/decode standard avro, as you know. But in your tests, you are just testing if encoding then decoding, all using avro4k, is compatible by itself. We should more see if the standard apache library is compatible with avro4k (and vice-versa).

Example of encoding test:
```kotlin
val rawAvroObject = getBytesFromApacheAvroLibrary(fooStringJava); // may be static bytes, since avro bytes should not move
val kotlinDecoded = Avro.default.openInputStream(serializer) {
         decodeFormat = AvroDecodeFormat.SingleObject(
            readerSchema = readerSchema,
            schemaStore = schemaStore
         )
      }.from(bytes).use { it.nextOrThrow() }
kotlinDecoded shouldBeEqual fooStringKotlin
```

Same for encoding.

This way you unit test encode & decode features, and also test compatibility from apache avro

Chuckame (changes requested):

src/test/kotlin/com/github/avrokotlin/avro4k/io/AvroSingleObjectInputStreamTest.kt
Chuckame: see https://github.com/avro-kotlin/avro4k/pull/142#discussion_r1304252441

src/test/kotlin/com/github/avrokotlin/avro4k/UseCaseSingleObjectEncodingTest.kt
jangalinski: removed it.

jangalinski:
Hi @Chuckame , I finally found the time to get back on this. 

What I did now: there is a String Constant containing: "[C3 01 CD 1D 19 01 C3 39 C6 61 06 62 61 72]"

This represents `FooString("bar")` in single object encoding ... written with the Java generated encoder.

The tests I rewrote now make sure that:

- we can decode the java-bytes using avro4k - compatibility for reading
- we can can encode kotlin to the same byte-hex (with the exclusion of the schema-fingerprint bytes, as these differ - we had to use different naming for the java-schema) -  compatibility with specification.
- we can encode kotlin and decode java. - compatibility for writing

I hope this is what you had in mind.

Chuckame:
Hey @jangalinski thanks it was something like that !

We are planning a v2 that will change a lot the codebase, and will make breaking changes, to make some room for all the new features and improvements, so merging your PR will add more work of adaptation. We will at least permit extension, at most implement this feature directly.

We won't close this PR unless the v2 implement it.

We are not forgetting this feature 😄 

Sorry for the very long delay.

Chuckame:
I'm closing it as it's going to be included in the v2

PR #152, opened 2023-08-04
properties with list of values

potfur:
Avro AVSC allows for custom annotations as shown [here](https://issues.apache.org/jira/browse/AVRO-3026).

Same functionality can be partially achieved by using `@AvroProp` annotation, but only for single value annotations. 

For cases - like the linked issue, where a list of values is needed, existing annotation is not enough.

Therefore I'm suggesting adding `@AvropProps("name", "valueA", "valueB", "etc")`

### DIFF
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
@@ -26,6 +26,7 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
          annotations.filterIsInstance<AvroAliases>().flatMap { it.value.toList() }
       }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
+   fun json(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
    fun default(): String? = annotations.filterIsInstance<AvroDefault>().firstOrNull()?.value
    fun enumDefault(): String? = annotations.filterIsInstance<AvroEnumDefault>().firstOrNull()?.value
 }
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
@@ -26,8 +26,7 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
          annotations.filterIsInstance<AvroAliases>().flatMap { it.value.toList() }
       }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
-   fun json(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
+   fun jsonProps(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
    fun default(): String? = annotations.filterIsInstance<AvroDefault>().firstOrNull()?.value
    fun enumDefault(): String? = annotations.filterIsInstance<AvroEnumDefault>().firstOrNull()?.value
 }
-
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
@@ -10,6 +10,10 @@ annotation class AvroProp(val key: String, val value: String)
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
+annotation class AvroJsonProp(val key: String, val jsonValue: String)
+
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class AvroNamespace(val value: String)
 
 @SerialInfo
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
@@ -62,7 +62,7 @@ annotation class AvroFixed(val size: Int)
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
-annotation class AvroDefault(val value: String)
+annotation class AvroDefault(@Language("json") val value: String)
 
 @SerialInfo
 @Target(AnnotationTarget.CLASS)
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
@@ -3,6 +3,7 @@ package com.github.avrokotlin.avro4k
 
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerialInfo
+import org.intellij.lang.annotations.Language
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
@@ -10,7 +11,7 @@ annotation class AvroProp(val key: String, val value: String)
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
-annotation class AvroJsonProp(val key: String, val jsonValue: String)
+annotation class AvroJsonProp(val key: String, @Language("json") val jsonValue: String)
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
--- a/src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
+++ b/src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
@@ -46,6 +46,17 @@ class AvroJsonPropSchemaTest : WordSpec() {
    data class AnnotatedProperties(
       @AvroJsonProp("guns", """["and", "roses"]""") val str: String,
       @AvroJsonProp("jean", """["michel", "jarre"]""") val long: Long,
+      @AvroJsonProp(
+         key = "object",
+         jsonValue = """{
+           "a": "foo",
+           "b": 200,
+           "c": true,
+           "d": null,
+           "e": { "e1": null, "e2": 429 },
+           "f": ["bar", 404, false, null, {}]
+         }"""
+      )
       val int: Int
    )
 
--- a/src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
+++ b/src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
@@ -1,0 +1,54 @@
+package com.github.avrokotlin.avro4k.schema
+
+import com.github.avrokotlin.avro4k.Avro
+import com.github.avrokotlin.avro4k.AvroJsonProp
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+import kotlinx.serialization.Serializable
+
+class AvroJsonPropSchemaTest : WordSpec() {
+
+   enum class Colours {
+      Red, Green, Blue
+   }
+
+   init {
+      "@AvroJsonProp" should {
+         "support props annotation on class" {
+
+            val expected = org.apache.avro.Schema.Parser()
+               .parse(javaClass.getResourceAsStream("/props_json_annotation_class.json"))
+            val schema = Avro.default.schema(TypeAnnotated.serializer())
+            schema.toString(true) shouldBe expected.toString(true)
+         }
+         "support props annotation on field" {
+
+            val expected = org.apache.avro.Schema.Parser()
+               .parse(javaClass.getResourceAsStream("/props_json_annotation_field.json"))
+            val schema = Avro.default.schema(AnnotatedProperties.serializer())
+            schema.toString(true) shouldBe expected.toString(true)
+         }
+         "support props annotations on enums" {
+
+            val expected = org.apache.avro.Schema.Parser()
+               .parse(javaClass.getResourceAsStream("/props_json_annotation_scala_enum.json"))
+            val schema = Avro.default.schema(EnumAnnotated.serializer())
+            schema.toString(true) shouldBe expected.toString(true)
+         }
+      }
+   }
+
+   @Serializable
+   @AvroJsonProp("guns", """["and", "roses"]""")
+   data class TypeAnnotated(val str: String)
+
+   @Serializable
+   data class AnnotatedProperties(
+      @AvroJsonProp("guns", """["and", "roses"]""") val str: String,
+      @AvroJsonProp("jean", """["michel", "jarre"]""") val long: Long,
+      val int: Int
+   )
+
+   @Serializable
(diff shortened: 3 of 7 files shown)


franckrasolo:
Hey @thake & @sksamuel 👋

Is there any chance this could be reviewed/approved at some point?

Chuckame:
Hey, I'll take a look

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
```diff
@@ -8,6 +8,10 @@ import kotlinx.serialization.SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class AvroProp(val key: String, val value: String)
 
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
+annotation class AvroProps(val key: String, vararg val value: String)
```
Chuckame, line 13: I think to improve the readability, it should be named `values` as it can contains one or more values, and it should be an array to distinguish key from values as shown in the example:
```kotlin
  @Serializable
  @AvroProps("cold", ["play", "and", "more"])
  data class TypeAnnotated(val str: String)
```
```suggestion
annotation class AvroProps(val key: String, vararg val values: Array<String>)
```

franckrasolo:
Thanks @Chuckame, much appreciated!

Chuckame (changes requested):

thake:
@potfur, thank you very much for the PR and sorry for the delay. 

Wouldn't it be enough to just change `@AvroProp` to be `@Repeatable`? This would raise the minimal kotlin level to 1.6 which is acceptable. Code:
```kotlin
@SerialInfo
@Repeatable
annotation class AvroProp(val key: String, val value: String)
```

Another option which does not require kotlin 1.6 is to define the new annotation `@AvroProp` like this:
```kotlin
@SerialInfo
annotation class AvroProps(vararg val props: AvroProp)

//Example usage
@Serializable
@AvroProps(
  AvroProp("cold", "play"),
  AvroProp("some", "thing")
)
data class TypeAnnotated(val str: String)
```

I'm in favor of the first solution as this is a classic example of why repeatable annotations have been introduced. But I would like to hear your opinion on it.

potfur:
The problem that I'm trying to solve is to be able to define 
```
"tag": ["val1", "val2"]
```

From what I know, repeating annotation will result in something like:
```
"tag1": "val1"
"tag2":"val2"
```

Chuckame:
With the comment of @thake it would result to something like:
```kotlin
@AvroProp("cold", "play")
@AvroProp("cold", "some")
@AvroProp("cold", "thing")
data class TypeAnnotated(val str: String)
```

Personally I don't really like it as it feels really verbose.

Chuckame:
Or, we could take into the middle: Just modify `AvroProp` to accept multiple values using vararg or array. (breaking change, but cleaner)

thake:
@potfur, thanks for fixing my confusion ;)

I'm totally with @Chuckame, the breaking change seems to be legit. @potfur can you also add the repeatable annotation? 
```kotlin
@SerialInfo
@Repeatable
annotation class AvroProp(val key: String, vararg val values: String)
```

potfur:
If breaking change is way to go - even better.

Chuckame:
Oh we missed something important @thake , that prevent us to have like a common `AvroProp` with single value and list value: We can have additional props like this `"key": "value"` or `"key": ["value"]`. How we can differentiate list and non-list added props ?

So new idea, we keep `AvroProp` having a single value (no vararg so it remains untouched) and another `AvroListProp` (same as the previously `AvroProps`) : 
```kotlin
annotation class AvroListProp(val key: String, values: Array<String>)
```

src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
```diff
@@ -25,7 +25,8 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
       }else{
          annotations.filterIsInstance<AvroAliases>().flatMap { it.value.toList() }
       }
-   fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
+   fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }.filterNot { it.second.isBlank() }
+   fun propsList(): List<Pair<String, List<String>>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.values.toList() }.filterNot { it.second.isEmpty() }
```
Chuckame, line 29: Little improvements:
- Allow empty list, as it could be a use case
- Don't convert to a List where it's just after used with values.forEach
```suggestion
   fun propsList(): List<Pair<String, Array<String>>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.values }
```

Chuckame (changes requested):

potfur:
Something like this (below) will work nicely, but has it's own limitations.

```kotlin
@SerialInfo
annotation class AvroProp(val key: String, val value: String = "", val values: Array<String> = [])
```

Mainly non-blank `value` must take precendence over `values`, otherwise it will end up in situation where same prop is declared twice and `AvroRuntimeException: Can't overwrite property: cold` is being thrown.

So maybe, let's go back to my initial proposal with minor improvements.

```kotlin
@SerialInfo
annotation class AvroProp(val key: String, val value: String)

@SerialInfo
annotation class AvroProps(val key: String, val values: Array<String>)
```

As for `AvroListProp` vs `AvroProps`, I think `AvroProps` is consistent with already existing annotations - `AvroAlias` & `AvroAliases`

Chuckame:
Let's go to your initial proposal yes. Just the name, I would prefer `AvroListProp` against `AvroProps`, since `AvroProps` is ambiguous, like "is it a list of props". `Aliases` is different where we should fix it and remove it prior to a `AvroAlias` with a vararg value. I'll create an issue to track it !

Chuckame (approved):

thake:
AvroProps seems to me a misleading name, as it is still just one property. I'll have a look into it tomorrow.

Chuckame:
@thake it has been changed to `AvroListProp` 👌

thake:
@thake, @potfur thanks for your patience. 
I rechecked the [avro spec](https://avro.apache.org/docs/1.11.1/specification/#schema-declaration) on this topic and while the current proposal would solve the discrepancy for array values, the specification explicitly allows additional properties of **any** type (i.e., boolean, number, object, array, string, null). Maybe we should consider this for this PR.

We have a similar situation with the `@AvroDefault` annotation. Maybe we can apply the same solution so that the API feels coherent.

Proposal:
```kotlin
@SerialInfo
@Repeatable
annotation class AvroJsonProp(
    val key: String,
    val jsonValue: String)

//Usage
data class TypeAnnotated(
    @AvroJsonProp("array", """["play","blub"]""")
    @AvroJsonProp("obj", """{"javascript":"object"}""")
    val str: String
)
```

Chuckame:
@thake I'm not really a fan of this, since json in strings are source of errors and ambiguous, and regarding performances it requires parsing the value each time with ser/deser. We will have to make a cache etc...

Why not just having a supplier class, as this example :
```kotlin
annotation class AvroGenericProp(
    val key: String,
    val valueSupplier: AvroPropSupplier)

interface AvroPropSupplier<T>{
fun getValue() : T
}

//usage
data class TypeAnnotated(
    @AvroGenericProp("array", CustomPropSupplier::class)
    val str: String
)

object CustomPropSupplier : AvroPropSupplier<CustomObject> {

**Comment by potfur** [comment]
I followed @thake suggestion, as it is more as Avro actually suggests with handling this case, and it also brings bit more flexibility.

**Comment by Chuckame** [comment]
I don't understand, avro suggest to handle all the types in additional props. We we continue using json while we have kotlin to write objects ? 🤔

File: src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt [hunk 3]
```diff
@@ -8,6 +8,10 @@ import kotlinx.serialization.SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class AvroProp(val key: String, val value: String)
 
+@SerialInfo
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
+annotation class AvroJsonProp(val key: String, val jsonValue: String)
```
Chuckame, line 13: We could use `@Language("json")` coming from jetbrains to validate json at least in jetbrains editors, what do you think ?

potfur:
@Chuckame I get your point, and for sure it would be better to use proper types.

On the other hand, the `AvroDefault` annotation already uses JSON strings, so `AvroJsonProp` is aligned and consistent what's already there (I like being consistent, even when it is suboptimal - gives some predictability and less cognitive load).

IMHO, doing improvements on types, and adding `Repeatable`, would be separate work - especially that forcing types would be a breaking change. So far, all is still backwards compatible - introduces new capability while preserving present.

Chuckame:
I have exactly the same concern about `AvroDefault` btw 🥲

Let's do AvroJsonProp, but it won't be that performant.

I'll create an issue to discuss about "type-safe" AvroDefault, AvroJsonProp and other jsons annotations.

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
franckrasolo: Yes, that would improve the development experience.

franckrasolo: I've just added it.

franckrasolo:
Given the backward-compatible nature of these changes, could we get a new release out once @thake has approved this PR?

src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
```diff
@@ -26,6 +26,7 @@ class AnnotationExtractor(private val annotations: List<Annotation>) {
          annotations.filterIsInstance<AvroAliases>().flatMap { it.value.toList() }
       }
    fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
+   fun json(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
```
thake, line 29:
```suggestion
   fun jsonProps(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
```

src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
```diff
@@ -0,0 +1,54 @@
+package com.github.avrokotlin.avro4k.schema
+
+import com.github.avrokotlin.avro4k.Avro
+import com.github.avrokotlin.avro4k.AvroJsonProp
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+import kotlinx.serialization.Serializable
+
+class AvroJsonPropSchemaTest : WordSpec() {
```
thake, line 9: Could you add some tests for object properties? Just to make sure that this is also correctly processed.

thake (changes requested):
Just some small remarks. After merging this, we'll release a new version.

thake:
@

> I have exactly the same concern about `AvroDefault` btw 🥲
> 
> Let's do AvroJsonProp, but it won't be that performant.
> 
> I'll create an issue to discuss about "type-safe" AvroDefault, AvroJsonProp and other jsons annotations.

It is a kind of trade-off between "ease of use" in combination with pre-existing schemas and type safety. Using JSON in the annotations makes it easy to model your kotlin code so that it matches an existing schema; you just copy and paste the value from the asvc schema file. 

Performance-wise, it shouldn't impact serialization and deserialization. The annotations' JSON is only interpreted whenever a schema is derived from the kotlin class. The schema is usually only derived once for every serialized class (responsibility of the user) and not on every serialization/deserialization.

src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
franckrasolo: Do you mean adding tests for a JSON value that is an object with properties?

franckrasolo: I ask because there's already the `@AvroJsonProp should support props annotation on field` test.

src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
franckrasolo: Function renamed

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
Chuckame: Can you add it to the `AvroDefault` annotation ? 👼

Chuckame:
@thake Since we'll have the `@Language` annotation to validate the content, it's an acceptable trade-off 🚀 

For the performances, If we use the `Avro` method `toRecord(serializer, obj): GenericRecord` then we regenerate the schema each time we serialize. Maybe we could add some doc to advice the use of `toRecord(serializer, schema, obj)` in an intensive serialization of the same schema.

Chuckame:
FYI, I'm playing with `@Repeatable` with kotlinx serialization, and it's only taking the latest annotated value, even if it should be already fixed. See https://github.com/Kotlin/kotlinx.serialization/issues/2099

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
franckrasolo: Done

src/test/kotlin/com/github/avrokotlin/avro4k/schema/AvroJsonPropSchemaTest.kt
thake: > Do you mean adding tests for a JSON value that is an object with properties?

Exactly

franckrasolo: I simply extended that test by annotating a property using a JSON object with heterogeneous property types.

thake (approved):
LGTM, thank you very much for your work!

franckrasolo:
@Chuckame / @thake Alas `@Language("json")` applied to `@AvroJsonProp` appears to only take effect in IntelliJ IDEA within the `avro-kotlin/avro4k` project itself, not when it's consumed from another project.

There are a number of issues such as [KTIJ-16874](https://youtrack.jetbrains.com/issue/KTIJ-16874/Custom-language-annotations-is-not-recognized-for-annotated-parameter) and [IDEA-173477](https://youtrack.jetbrains.com/issue/IDEA-173477/Language-meta-annotation-injection-does-not-work-for-a-method-argument) related to the processing of `@Language` as a meta-annotation. You may also want to vote them up.

Chuckame:
@franckrasolo as I can read, both are related not to our case: we are directly using this language annotation on the value parameter. Those issues are related to meta annotations, implying that annotations (like `@Language`) are annotating an annotation `A`, where this annotation `A` is used on a parameter, so all the annotations on `A` will be applied to the parameter.

Also tested AvroProp on my computer and the json language injection is not working. After a quick check, tldr: replace json by JSON.
Longer story: language injection is done by intellij, and we can find all the languages in Parameters > Editor > Language injections. We should set a value found inside the Language column, where we can only find JSON and not json, maybe it's case sensitive! I'll do the pr

franckrasolo:
Yes, not an exact match, but the number of issues makes me wonder how stable the processing of `@Language` is across a wide range of use cases: host languages x injected languages x IDE flavours, and so forth.

PR #156, opened 2023-08-25
feat(aliases): Merge AvroAliases annotation to AvroAlias

### LINKED ISSUE #155: Remove Aliases, prior to Alias with vararg
Currently we have `Alias` for single alias and `Aliases` for multiple aliases where we could simplify to a single `Alias` with a vararg value:
```kotlin
annotation class Alias(vararg values: String)
```

Chuckame:
- Now, to have multiple aliases, just annotate like `@AvroAlias("alias1", "alias2", "alias3")` as it is now a vararg parameter
- updated to kotlin 1.9 to have `@Repeatable` working
- deprecated `@AvroAliases` in favor of `@AvroAlias`
- Closes #155

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,7 +50,7 @@ dependencies {
 tasks.withType<KotlinCompile>().configureEach {
    kotlinOptions.jvmTarget = "1.8"
    kotlinOptions.apiVersion = "1.5"
-   kotlinOptions.languageVersion = "1.5"   
+   kotlinOptions.languageVersion = "1.5"
    kotlinOptions.freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
 }
 java {
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
@@ -35,9 +35,10 @@ annotation class AvroDoc(val value: String)
 
 @SerialInfo
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
-annotation class AvroAlias(val value: String)
+annotation class AvroAlias(vararg val value: String)
 
 @SerialInfo
+@Deprecated(message = "Will be removed in the next major release", replaceWith = ReplaceWith("@AvroAlias(alias1, alias2)"))
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class AvroAliases(val value: Array<String>)
 
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
@@ -2,6 +2,7 @@ package com.github.avrokotlin.avro4k
 
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.descriptors.SerialDescriptor
+
 @ExperimentalSerializationApi
 class AnnotationExtractor(private val annotations: List<Annotation>) {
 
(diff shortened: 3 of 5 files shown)


Chuckame:
@thake Is it ok for you, and do you feel comfortable by changing from `AvroAlias(val value: String)` to `AvroAlias(vararg val value: String)` with only deploying a new feature (still `1.x`) and not a breaking change (in `2.x`) ?

thake:
@Chuckame, I'm comfortable with this approach. It should be highlighted in the release notes, though.

build.gradle.kts
```diff
@@ -49,8 +49,8 @@ dependencies {
 
 tasks.withType<KotlinCompile>().configureEach {
    kotlinOptions.jvmTarget = "1.8"
-   kotlinOptions.apiVersion = "1.5"
-   kotlinOptions.languageVersion = "1.5"   
+   kotlinOptions.apiVersion = "1.9"
```
thake, line 52: Is it required to increase the apiVersion? It requires all of our users to update their codebase to kotlin 1.9. This seems to be a big change for a small cosmetic API change.

build.gradle.kts
```diff
@@ -49,8 +49,8 @@ dependencies {
...
+   kotlinOptions.languageVersion = "1.9"
```
thake, line 53: see above

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
```diff
@@ -28,11 +28,13 @@ annotation class AvroInline
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class AvroDoc(val value: String)
 
+@Repeatable
```
thake, line 31: Does it make sense to have it @Repeatable here? I would rather remove it, as the user has the possibility to use the vararg.

thake (changes requested):

build.gradle.kts
Chuckame: Now we don't want the `@Repeatable`. But if we want it for another annotation, then we will need kotlin 1.9. Previous versions has been tested and we can only see the last (or first) repeated annotation

Chuckame: And also, keeping a `-4` version is the open door to migration surprises, maybe we should keep a support to `-1` or `-2` versions no ? By the way I'm not totally comfortable about kotlinVersion (libraries) vs apiVersion vs languageVersion, what are the real impacts on users. I'll take a look

Chuckame: And also, currently I'm working with the latest version of kotlinx serialization that is on kotlin 1.8 or 1.9. If we want a full power library, then we will need to migrate

buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -15,7 +15,7 @@ object Libs {
    }
 
    object Kotlinx {
-      private const val version = "1.5.0"
```
Chuckame, line 18: From your other comment:
> It requires all of our users to update their codebase to kotlin 1.9

Here we used version `1.5.0` that were using kotlin 1.8 ([see the release notes](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0))

build.gradle.kts
Chuckame: back to 1.5

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
Chuckame: no repeatable anymore

Chuckame:
@thake ok for you ?

thake (approved):

Chuckame:
Let's m

Chuckame:
Let's m

Issue #151, opened 2023-07-25
Performance bottleneck at RecordNaming


twinprime:
Hi, I am not sure if this is a known problem.

I was doing a simple benchmark of Avro4k and discovered that when Unions are involved, e.g. a sealed interface, a lot of CPU time (like 30% for my case) is spent on `RecordNaming.invoke(String, List<Annotation>, NamingStrategy)`. I used VisualVM to verify.

This is one of the thing that makes Avro4K much slower than jackson-dataformat-avro. Its 2-3x slower for my test case.

Chuckame:
Thanks for the benchmark. Can you put at least a screenshot of visualvm of CPU Time tab ? or at most the file ?

We know that there is still reflection that could be replaced, and a lot of re-computation on some cases.

twinprime:
No problem, I uploaded the src and a screen shot here

[https://github.com/twinprime/avro-benchmark](https://github.com/twinprime/avro-benchmark)

Chuckame:
Thanks. I'll try to fix this

Chuckame:
I'm currently working on #148 that should improve the performances, I'll let you informed

thake:
A huge performance improvement is planned for version 2. See https://github.com/avro-kotlin/avro4k/issues/160#issuecomment-1711616190

Chuckame:
Duplicate of #199

Chuckame:
FYI @twinprime I just move your project to avro4k to allow us benchmarking easily for the next releases. See #208

Issue #147, opened 2023-06-16
Native image issue


rafalg:
Hi,

Trying to run my app as an native image. However get this issue: 

`java.lang.NoSuchFieldException: original
	at java.base@17.0.5/java.lang.Class.getDeclaredField(DynamicHub.java:961)
	at com.github.avrokotlin.avro4k.schema.SchemaForKt.schemaFor(SchemaFor.kt:170)
`
Is it a must for avro4k to use reflection?

Chuckame:
Hello,
It's a good question.
In fact, we should not use reflection since the core of using kotlinx serialization is to not have reflection. This line is about a tricks that unwrap a nullable field serializer to maybe get an internal avro serializer.

We are currently re thinking about the schema generation by fully using kotlinx serialization features.

It won't be that quick, so the quick answer is yes, and let's see in the future to not have reflection anymore.

Chuckame:
See https://github.com/avro-kotlin/avro4k/issues/148

Chuckame:
Still using reflection with `AvroSchema` and `AvroLogicalType`. Waiting for the next kotlinx-serialization release that will include this PR to allow accessing the descriptor without reflection: https://github.com/Kotlin/kotlinx.serialization/pull/2633

Chuckame:
Duplicate of #199

Issue #116, opened 2021-09-20
SerializationException when using MapSerializer as root serializer


FranziskusW:
I am using `Avro.default.toRecord(serializer, data)` with a `MapSerializer(String.serializer(), String.serializer()`. As it returns `kotlinx.serialization.SerializationException: Unsupported root element passed to root record encoder` I investigated and found the following snippet in 

```
override fun beginStructure(descriptor: SerialDescriptor): CompositeEncoder {
      return when (descriptor.kind) {
         is StructureKind.CLASS -> RecordEncoder(schema, serializersModule, callback)
         is PolymorphicKind -> UnionEncoder(schema,serializersModule,callback)
         else -> throw SerializationException("Unsupported root element passed to root record encoder")
      }
}
```

It seems that `StructureKind.CLASS` is the culprit. It contains an object of that type, but the kotlin `is`-comparator expects a type.

FranziskusW:
As a minimal example for reproduction, you can try this:
```
import kotlinx.serialization.SerializationException
 import kotlinx.serialization.builtins.MapSerializer
 import kotlinx.serialization.descriptors.PolymorphicKind
 import kotlinx.serialization.descriptors.StructureKind
 
 import kotlinx.serialization.builtins.serializer
 println(
     when (MapSerializer(String.serializer(), String.serializer()).descriptor.kind) {
         is StructureKind -> "StructureKind.CLASS"
         is PolymorphicKind -> "PolymorphicKind"
         else -> throw SerializationException("Unsupported root element passed to root record encoder")
     }
 )
```

FranziskusW:
Is this a restriction of AVro itself? Not allowing a Map as root element?

thake:
Currently avro4k only supports data classes as roots. But it should be possible to also support all other types. Do you want to give it a try and provide a pr?

Chuckame:
Closing stale issue. By the way, we are planning to handle non-record values

Chuckame:
Closing this issue, as it will be covered by #187 and #114

PR #161, opened 2023-09-01
feat: encode/decode anything as root value

### LINKED ISSUE #80: encodeToByteArray() fails when applied with primitives
for example:
```
Avro.default.encodeToByteArray(String.serializer(), "foo"))
```
triggers
```
Exception in thread "main" kotlinx.serialization.SerializationException: Non-serializable class kotlin.String is not supported by class com.sksamuel.avro4k.encoder.RootRecordEncoder encoder
	at kotlinx.serialization.encoding.AbstractEncoder.encodeValue(AbstractEncoder.kt:36)
	at kotlinx.serialization.encoding.AbstractEncoder.encodeString(AbstractEncoder.kt:50)
	at kotlinx.serialization.internal.StringSerializer.serialize(Primitives.kt:139)
	at kotlinx.serialization.internal.StringSerializer.serialize(Primitives.kt:137)
	at kotlinx.serialization.encoding.Encoder$DefaultImpls.encodeSerializableValue(Encoding.kt:259)
	at kotlinx.serialization.encoding.AbstractEncoder.encodeSerializableValue(AbstractEncoder.kt:18)
	at com.sksamuel.avro4k.Avro.toRecord(Avro.kt:220)
	at com.sksamuel.avro4k.Avro$openOutputStream$builder$1$1.invoke(Avro.kt:199)
	at com.sksamuel.avro4k.Avro$openOutputStream$builder$1$1.invoke(Avro.kt:104)
	at com.sksamuel.avro4k.io.AvroDataOutputStream.write(AvroDataOutputStream.kt:44)
	at com.sksamuel.avro4k.Avro.encodeToByteArray(Avro.kt:179)
	at io.infinitic.common.MainKt.main(main.kt:18)
```

### LINKED ISSUE #114: Allow non-string while string-able values as map keys
Hey folks, this library looks pretty good so far, but I'm running into one awkwardness around Map properties keyed by something other than `String`.

When the key type is a `data class` in my own codebase, I can annotate it with a serializer with a primitive descriptor, and avro4k is happy to use that. But for an external type, I'd like to be able to have properties like `Map<@Contextual SomethingExternal, Foo>` and have avro4k look up the serializer for `SomethingExternal` dynamically from context, but it doesn't seem to do so.

I can set `@file:UseSerializers` on the file, but I'd rather use `@Contextual`. Thanks!

### LINKED ISSUE #131: Value classes not working (feature request ?)
Hey,

Is there a way to have value classes ? (previously called inlined classes)

The goal is to wrap primitive types (string, int, in our example) inside a value class, to be used with a primitive avro schema.

error:
```
Non-serializable class kotlin.String is not supported by class com.github.avrokotlin.avro4k.encoder.RootRecordEncoder encoder
```

To reproduce:
```kotlin
@JvmInline
@Serializable
value class StringValueClass(val value: String)

val kotlinSerializer = serializer<StringValueClass>() // working well
val avroSchema = Avro.default.schema(kotlinSerializer) // working well, creates a "string" schema
Avro.default.toRecord(kotlinSerializer,avroSchema, StringValueClass("awesome"))
```

I can help if you want, since this feature should be awesome for handling primitive keys while having a readable code (instead of ask "hey, what's inside the string key ? hexadecimal ? the name ? from whhich other field it comes ?" 😄 )

### LINKED ISSUE #148: Use kotlinx serializer native features
This issue will be the follow up to assist the long way of using native kotlinx serialization features like `@Contextual`, `@SerialName`, `serializerModules` and much more to unlock native features like value classes, serialization modules declaration, custom serializers.

This work would touch to the schema generation and the encoding/decoding codebase, while we will try to not touch the user entrypoint `Avro`.

Chuckame:
This PR needed a big refactor to handle easily 
- Now able to encode/decode anything as root value (not only records) such as maps, lists, primitives, value classes... (closes #80 and #116) 
- Handle value classes and contextual (that encodes as string) for map keys (closes #114) 
- No reflection at all, only kotlinx serializers & descriptors and improve naming by using `@SerialName` for namespace and record name when contains dots (closes #148 and #151)
- Handle unions on kotlin primitive fields (by example: if the given field schema is a union with an int, while calling `encodeInt`, then it's working)
- Encode by default char (code), byte and short to an avro int (was implicit before for byte and char) 
- value classes 100% handled (so AvroInline removed as it prevent native value class serializers to work well) closes #131 
- fields of kotlin classes are not forced to be in same order of avro schema fields
- unified and refactored encoders, decoders, schema generation and `Avro` user entry point. Pay attention of deprecations 

#147 partially done (encoder & decoder only). We still have one use case of reflection in schema generation, but will be done into another PR


Chuckame:
Hey @thake , I'll leave some comments to let you focus or understand some specificities. I tried my best to touch on encoders classes, then decoders classes, then schemas classes.

I don't really know how to split this mess since a lot of things permits the new features together.

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/RecordEncoder.kt
```diff
@@ -1,110 +1,49 @@
...
+
+    override val currentUnresolvedSchema: Schema get() = currentField.schema()
+
+    override fun encodeNativeValue(value: Any?) {
+        builder.set(currentField, value)
+    }
+
+    override fun shouldEncodeElement(descriptor: SerialDescriptor, index: Int): Boolean {
+        val fieldName = descriptor.getElementAvroName(configuration.namingStrategy, index)
+        val foundField = recordSchema.getField(fieldName)
```
Chuckame, line 38: > fields of kotlin classes are not forced to be in same order of avro schema fields

Regarding performances, accessing an array from index (kotlin fields must be in same order as avro schema) or accessing a hashtable with a string hash, is it different ? I yes, we can detect good order during the beginStructure by returning a `WellSortedRecordEncoder` or `NotWellSortedRecordEncoder`

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/MapEncoder.kt
```diff
@@ -1,63 +1,55 @@
...
+        if (currentKey == null) {
+            currentKey = if (value is Utf8) {
+                value
+            } else {
+                Utf8(value?.toString())
+            }
+        } else {
+            finalizeMapEntry(value)
+        }
+    }
```
Chuckame, line 39: > Maps can now have a key not only with string type

value class of String, primitives, null, and maybe complex objects can be set as key, but with the detail were we use now `value.toString()`

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/PolymorphicEncoder.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.CompositeEncoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
+
+
+@ExperimentalSerializationApi
+abstract class PolymorphicEncoder(
```
Chuckame, line 13: Previously was `UnionEncoder`, but since we can now encode a string with schema `["string", "int"]` thanks to the new resolve system, we can also encode polymorphic types but with an instance that only matches the non-union schema (e.g. we can write a polymorphic type `Parent` where children are `Child1` and `Child2`, with the non-union schema `{type: Child1}` only if the given instance is type of Child1. If it's Child2, then it's failing.).

It's totally outside the avro world, and could be used by any other format. This class put clear methods `encodeSerialName` and `encodeValue` instead of respectively unclear `encodeStringElement` and `encodeSerializableElement` so adds lisibility when using polymorphic encoder.

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/RecordEncoder.kt
```diff
@@ -1,110 +1,49 @@
...
+        if (foundField != null) {
+            currentField = foundField
+            return true
+        }
+        throw AvroRuntimeException("Missing field $fieldName in schema $recordSchema. This means that the kotlin class is containing a field that is not present in the given schema. Try use another schema or annotate the original class field ${descriptor.getElementName(index)} with @kotlinx.serialization.Transient to skip it")
```
Chuckame, line 43: We could also add a new feature of ignoring kotlin fields that are not specified in avro schema !

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/RootRecordEncoder.kt
```diff
@@ -1,30 +0,0 @@
...
-import kotlinx.serialization.descriptors.PolymorphicKind
-import kotlinx.serialization.descriptors.SerialDescriptor
-import kotlinx.serialization.descriptors.StructureKind
-import kotlinx.serialization.encoding.AbstractEncoder
-import kotlinx.serialization.encoding.CompositeEncoder
-import kotlinx.serialization.modules.SerializersModule
-import org.apache.avro.Schema
-
-@ExperimentalSerializationApi
-class RootRecordEncoder(private val schema: Schema,
```
Chuckame, line 15: Replaced by `GenericAvroEncoder`

src/main/kotlin/com/github/avrokotlin/avro4k/AnnotationExtractor.kt
```diff
@@ -1,28 +1,65 @@
...
-   fun name(): String? = annotations.filterIsInstance<AvroName>().firstOrNull()?.value
-   fun valueType(): Boolean = annotations.filterIsInstance<AvroInline>().isNotEmpty()
-   fun doc(): String? = annotations.filterIsInstance<AvroDoc>().firstOrNull()?.value
-   fun aliases(): List<String> = (annotations.firstNotNullOfOrNull { it as? AvroAlias }?.value ?: emptyArray()).asList() + (annotations.firstNotNullOfOrNull {it as? AvroAliases}?.value ?: emptyArray())
-   fun props(): List<Pair<String, String>> = annotations.filterIsInstance<AvroProp>().map { it.key to it.value }
-   fun jsonProps(): List<Pair<String, String>> = annotations.filterIsInstance<AvroJsonProp>().map { it.key to it.jsonValue }
-   fun default(): String? = annotations.filterIsInstance<AvroDefault>().firstOrNull()?.value
-   fun enumDefault(): String? = annotations.filterIsInstance<AvroEnumDefault>().firstOrNull()?.value
+@ExperimentalSerializationApi
+internal fun SerialDescriptor.getElementAvroName(namingStrategy: NamingStrategy, index: Int): String {
```
Chuckame, line 54: I prefer `descriptor.getElementAvroName(namingStrategy, index)` than old java helper class fashion `AnnotationExtractor(descriptor.getElementAnnotations(index)).avroName/namespace...`.

I would do the same for all other methods here to have great extensions

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -19,262 +19,314 @@ import kotlinx.serialization.SerializationStrategy
...
+            encode(serializer, schema(serializer), value)
+
+    fun <T> encode(serializer: SerializationStrategy<T>,
+                   schema: Schema,
+                   value: T): Any? {
+        val encoder = GenericAvroEncoder(schema, serializersModule, configuration)
+        encoder.resolveSchema(serializer.descriptor, value == null)
+        serializer.serialize(encoder, value)
+        return encoder.encodedValue
+    }
```
Chuckame, line 283: What has changed in this file is only the handling of all value types in root, nothing else

src/main/kotlin/com/github/avrokotlin/avro4k/RecordNaming.kt
```diff
@@ -7,58 +7,47 @@ import kotlinx.serialization.descriptors.SerialDescriptor
...
 
    companion object {
-      operator fun invoke(name: String, annotations: List<Annotation>, namingStrategy: NamingStrategy): RecordNaming {
-         val className = name
-            .replace(".<init>", "")
-            .replace(".<anonymous>", "")
+      operator fun invoke(name: String, annotations: List<Annotation>, namingStrategy: NamingStrategy = DefaultNamingStrategy): RecordNaming {
+         val className = if (name.contains('<')) name
+            .replaceFirst(".<init>", "")
+            .replaceFirst(".<anonymous>", "") else name
```
Chuckame, line 36: kotlinx serialization maybe did this in the past, but it's not possible currently to have this kind of name

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
```diff
@@ -25,10 +25,6 @@ annotation class AvroName(val value: String)
 @Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS)
 annotation class ScalePrecision(val scale: Int, val precision: Int)
 
-@SerialInfo
-@Target(AnnotationTarget.CLASS)
-annotation class AvroInline
```
Chuckame, line 30: Now it's value classes

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/AbstractAvroEncoder.kt
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.avro.Schema
+import org.apache.avro.generic.GenericData
+import org.apache.avro.generic.GenericData.EnumSymbol
+import org.apache.avro.generic.GenericFixed
+import java.math.BigDecimal
+import java.nio.ByteBuffer
+
+
+@OptIn(ExperimentalSerializationApi::class)
+abstract class AbstractAvroEncoder : Encoder, NativeAvroEncoder {
```
Chuckame, line 27: Note here, we are using Encoder instead of AbstractEncoder, to split only value encoding (GenericAvroEncoder) and structure encoding (map, list and record). AbstractEncoder implements both Encoder and CompositeEncoder

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/AvroStructureEncoder.kt
```diff
@@ -0,0 +1,95 @@
...
+    open fun shouldEncodeElement(descriptor: SerialDescriptor, index: Int): Boolean =
+            true
+
+    //region composite methods
+    final override fun encodeBooleanElement(descriptor: SerialDescriptor, index: Int, value: Boolean) {
+        if (shouldEncodeElement(descriptor, index)) {
+            resolveElementSchema(descriptor, index, false)
+            encodeBoolean(value)
+        }
+    }
```
Chuckame, line 20: Now, for each element, we resolve the schema if it's a union. This permits to encode a string even if the union is `["null", "string", "int", {complex object I don't know why...}]`. Great for generic schemas but specific classes, for free

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/GenericAvroEncoder.kt
```diff
@@ -0,0 +1,22 @@
...
+) : AbstractAvroEncoder() {
+    var encodedValue: Any? = null
+        private set
+
+    override fun encodeNativeValue(value: Any?) {
+        encodedValue = value
+    }
+
+    override val currentUnresolvedSchema: Schema = schema
+}
```
Chuckame, line 22: Why I prefer to not split the not-only-records in root ? This class, it seems easy, but needs the rest to work 😄

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/ListEncoder.kt
```diff
@@ -1,78 +1,30 @@
...
+class ListEncoder(
+        listSchema: Schema,
+        size: Int,
+        override val serializersModule: SerializersModule,
+        override val configuration: AvroConfiguration,
+        private val callback: (GenericArray<Any?>) -> Unit,
+) : AvroStructureEncoder() {
+    override val currentUnresolvedSchema: Schema = listSchema.elementType
+
+    private val list = GenericData.Array<Any?>(size, listSchema)
```
Chuckame, line 21: Little perf improvement: Since we use directly `GenericData.Array`, apache avro library won't need to copy the list content to a GenericData.Array, we do it directly, and with an array size to not grow the array for nothing

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/MapEncoder.kt
```diff
@@ -1,63 +1,55 @@
...
-
-   override fun fieldSchema(): Schema = valueSchema
+class MapEncoder(
+        mapSchema: Schema,
+        mapSize: Int,
+        override val serializersModule: SerializersModule,
+        override val configuration: AvroConfiguration,
+        private val callback: (Map<Utf8, *>) -> Unit,
+) : AvroStructureEncoder() {
+    private val map = HashMap<Utf8, Any?>(Math.multiplyExact(mapSize, 2)) // times 2 to prevent the map growing
```
Chuckame, line 19: free perf improvement, avoid already-known growings by setting the hash map size

Chuckame:
Next step: add some tests for the new features, and implement decode stuff

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/MapEncoder.kt
Chuckame: keeping only strings

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/RecordEncoder.kt
Chuckame: done

Chuckame:
Cool @thake ! By the way, the work I've did on the encoder/decoder classes will help a lot to really easily encode and decode without using apache library, I would advice you to have a look, it's a common structure when doing kotlinx serialization 🚀

Chuckame:
@thake It's ready for review. FYI, I will scope this PR only for encoders & decoders about the refactoring. But the main features are covered (read the PR description, there is all the fixes). Schemas are already handling everything as root, but with some reflection and weird stuff with custom serial descriptor. I have the solution (WIP), I'll create another PR !

Hope you'll appreciate this work 😄

src/main/kotlin/com/github/avrokotlin/avro4k/schema/UnionSchemaFor.kt
```diff
@@ -17,7 +17,7 @@ class UnionSchemaFor(
 ) : SchemaFor {
     override fun schema(): Schema {        
         val leafSerialDescriptors =
-            descriptor.possibleSerializationSubclasses(serializersModule).sortedBy { it.serialName }
+            descriptor.possibleSerializationSubclasses(serializersModule).sortedBy { it.serialName }.toList()
```
thake, line 20: I would move the toList() in the next call (end of line 23) in order to save one iteration over all sequence elements.

src/main/kotlin/com/github/avrokotlin/avro4k/schema/namingStrategy.kt
```diff
@@ -1,22 +1,31 @@
...
 object SnakeCaseNamingStrategy : NamingStrategy {
-   override fun to(name: String): String = name.fold(StringBuilder()) { sb, c ->
-      if (c.isUpperCase())
-         sb.append('_').append(c.lowercase())
-      else
-         sb.append(c.lowercase())
-   }.toString()
+    override fun to(name: String): String {
+        var previousWasUppercase = true
+        return name.foldIndexed(StringBuilder(name.length)) { index, sb, c ->
```
thake, line 18: use fold as the index is not used

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/AvroSerializer.kt
```diff
@@ -1,39 +1,22 @@
...
-//      val subschema = when (schema.type) {
-//         Schema.Type.UNION -> schema.extractNonNull()
-//         else -> schema
-//      }
-      return decodeAvroValue(schema, decoder)
+       return decodeAvroValue((decoder as NativeAvroDecoder).currentSchema, decoder)
    }
 
-   abstract fun encodeAvroValue(schema: Schema, encoder: ExtendedEncoder, obj: T)
+   abstract fun encodeAvroValue(schema: Schema, encoder: NativeAvroEncoder, obj: T)
```
thake, line 19: This is a breaking change with the outer API. If possible, I would not introduce it with this patch. There will be an upcoming interface change if we start our way to multiplatform. So let us start a discussion on how the API should look like taking the upcoming change to natively encode/decode into account. That way, we only change the API once.

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/BigDecimalSerializer.kt
```diff
@@ -16,68 +15,58 @@ import org.apache.avro.LogicalTypes
...
+/**
+ * A BigDecimal is serialized in 3 ways:
+ * - as a [String] when schema type is [Schema.Type.STRING] (scale and precision have not effect here)
+ * - as a [ByteBuffer] when schema type is [Schema.Type.BYTES] (see [Conversions.DecimalConversion.toBytes])
+ * - as a [GenericFixed] when schema type is [Schema.Type.FIXED] (see [Conversions.DecimalConversion.toFixed])
+ */
 @OptIn(ExperimentalSerializationApi::class)
-@Serializer(forClass = BigDecimal::class)
-class BigDecimalSerializer : AvroSerializer<BigDecimal>() {
+object BigDecimalSerializer : AvroSerializer<BigDecimal>() {
```
thake, line 29: I never thought about that :) :rocket:

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/BigDecimalSerializer.kt
```diff
@@ -16,68 +15,58 @@ import org.apache.avro.LogicalTypes
...
 import kotlin.reflect.jvm.jvmName
 
+/**
+ * A BigDecimal is serialized in 3 ways:
+ * - as a [String] when schema type is [Schema.Type.STRING] (scale and precision have not effect here)
+ * - as a [ByteBuffer] when schema type is [Schema.Type.BYTES] (see [Conversions.DecimalConversion.toBytes])
+ * - as a [GenericFixed] when schema type is [Schema.Type.FIXED] (see [Conversions.DecimalConversion.toFixed])
+ */
 @OptIn(ExperimentalSerializationApi::class)
-@Serializer(forClass = BigDecimal::class)
```
thake, line 26: I always thought this would be mandatory, but it seems that I've not correctly understood the documentation. Thanks for removing it!

src/test/kotlin/com/github/avrokotlin/avro4k/stuff.kt
```diff
@@ -0,0 +1,26 @@
+package com.github.avrokotlin.avro4k
```
thake, line 1: Maybe this file is better named utils.kt?

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -15,266 +15,322 @@ import kotlinx.serialization.BinaryFormat
...
-      get() {
-         return when(format){
-            is AvroFormat.JsonFormat -> {
-               val wschema = writerSchema ?: error("Writer schema needs to be supplied for Json format")
-               AvroDecodeFormat.Json(wschema, readerSchema?:wschema)
-            }
-            is AvroFormat.BinaryFormat -> {
-               val wschema = writerSchema ?: error("Writer schema needs to be supplied for Binary format")
-               AvroDecodeFormat.Binary(wschema, readerSchema?:wschema)
+    private val converter: (Any) -> T
```
thake, line 39: It is really hard to review this patch with all the reformating ...

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
```diff
@@ -15,266 +15,322 @@ import kotlinx.serialization.BinaryFormat
...
+
+inline fun <reified T> Avro.encodeToGenericData(value: T): Any? {
+    val serializer = serializersModule.serializer<T>()
+    return encodeToGenericData(serializer, schema(serializer.descriptor), value)
+}
+
+fun <T> Avro.encodeToGenericData(serializer: SerializationStrategy<T>, value: T): Any? =
+    encodeToGenericData(serializer, schema(serializer.descriptor), value)
+
+inline fun <reified T> Avro.encodeToGenericData(schema: Schema, value: T) =
```
thake, line 324: how about merging it with the other reified encodeToGenericData method?
```kotlin
inline fun <reified T> Avro.encodeToGenericData(value: T, serializer: SerializationStrategy<T> = serializersModule.serializer<T>(), schema: Schema = schema(serializer.descriptor)): Any? {
    return encodeToGenericData(value, serializer, schema)
}
```
I would change the signature of the encodeToGenericData method in the same way that the parameters have the following order: value, serializer, schema.

That way users can use the API in the following way:
```kotlin
Avro.default.encodeToGenericData("blu", serializer, schema)
Avro.default.encodeToGenericData("blu")
```

src/main/kotlin/com/github/avrokotlin/avro4k/annotations.kt
thake: As this is a breaking change, I would add a deprecation tag at least with one release before removing it.

thake (changes requested):
Thanks for your great work! It took some time to get through all the changes. Sorry for letting you wait.

Maybe we should start tagging our API parts of the library in order to get some tool support for detecting breaking changes (see https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html). What do you think?

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/AvroStructureEncoder.kt
```diff
@@ -0,0 +1,95 @@
+package com.github.avrokotlin.avro4k.encoder
+
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.CompositeEncoder
+import kotlinx.serialization.encoding.Encoder
+
+@ExperimentalSerializationApi
+abstract class AvroStructureEncoder : AbstractAvroEncoder(), CompositeEncoder {
```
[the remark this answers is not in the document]
Chuckame: https://github.com/avro-kotlin/avro4k/pull/161#discussion_r1313186720

Just the entry point is not a composite encoder. Regarding complexity, the other main reason of AvroStructureEncoder is to intercept encodeXxxElement to resolve schema if needed. But we can also resolve the type for each encodeXx methods, or not resolve schema at all, but then again, we are assuming that the schema absolutely come from avro4k

src/main/kotlin/com/github/avrokotlin/avro4k/encoder/MapEncoder.kt
```diff
@@ -1,63 +1,71 @@
...
+    override fun endStructure(descriptor: SerialDescriptor) {
+        callback(map)
+    }
+
+    override val currentUnresolvedSchema: Schema = mapSchema.valueType
 }
+
+private val STRING_SCHEMA = Schema.create(Schema.Type.STRING)
+
+private fun isKeyElement(index: Int) = index % 2 == 0
```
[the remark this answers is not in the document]
Chuckame: It just adds readability about modulo stuff. I can add the `inline` keyword but it would not change anything regarding performances

src/main/kotlin/com/github/avrokotlin/avro4k/schema/UnionSchemaFor.kt
Chuckame: good catch

src/main/kotlin/com/github/avrokotlin/avro4k/schema/namingStrategy.kt
Chuckame: good catch

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/AvroSerializer.kt
Chuckame: True, but there is a lot of over-interfaces with non explicit namings, so yes let's discuss about it 😄 i'll rollback the change about the naming

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/BigDecimalSerializer.kt
Chuckame: And I'm also working on how to remove the use of our custom SerialDescriptor by using annotations, a PR is coming

src/test/kotlin/com/github/avrokotlin/avro4k/stuff.kt
Chuckame: hahaha oops 😆 i'll do it

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
Chuckame: btw, i've put space formatting to 4 spaces, because a lot of files were using 4 spaces for tabs (but the Avro class only 3, it's a mess)

src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt
Chuckame: Good point, I just reused the same as before if I remember

Chuckame: I'll put the serialization strategy at the end as it is not required in most cases, so in this order: value, schema, serializer

Chuckame: I can't, so let's go in your order

Chuckame: The issue is that we have this error now: Extension is shadowed by a member. I don't have easy solution.

src/main/kotlin/com/github/avrokotlin/avro4k/SerialDescriptor.kt
```diff
@@ -1,20 +1,11 @@
 package com.github.avrokotlin.avro4k
 
 import kotlinx.serialization.ExperimentalSerializationApi
```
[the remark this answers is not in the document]
Chuckame: good catch

thake:
Just to let you know about how I work with GitHub PRs so that we don't have a misunderstanding: If I have submitted a review result, I won't look at the PR again until you "Re-request review".

So whenever you feel ready, just re-request my review :rocket:

Chuckame:
Ok, good point 👍 I'm currently taking a pause about avro4k, I'll continue working on it in 1-2 weeks 😄

Chuckame:
There is too much changes, I'll do some other atomic PRs instead of this big-bang PR

thake:
Let's consolidate our efforts on this topic, because https://github.com/avro-kotlin/avro4k/tree/way-to-multiplatform also includes a fix for this problem.

Chuckame:
I think first we should work to improve performance and readability as currently it's very difficult to add feature (kotlinx serialization framework not really well used), then maybe removing the avro lib dependency, and thinking about multiplatform.

PR #167, opened 2023-09-08
WIP Way to multiplatform

Chuckame:


Chuckame:
@thake can you provide how you did the benchmark ? Then I would also like to do it on my branch 😄

Issue #160, opened 2023-08-29
Bypass generic data step and serialize directly to binary


Chuckame:
Currently, the library is just interfacing as it can to the official apache avro library using GenericRecord and other GenericData stuff.

On avro4k codebase simplicity, it seems easy while it's not really since the codebase is doing a lot of adaptation to let the GenericData happy with the generic stuff we generated. 

On performance side, we are checking, converting and adapting data to fit GenericData stuff, and then this adapted stuff is also tried to be re-adapted and checked, to be then serialized to binary. Also, all is runtime specific while kotlinx serialization is mainly prepared at build time (except contextual). 

On user side, each user that just want to do avro format (and not directly use the generated GenericRecords) will have to call the apache avro library to serialize it to binary or json, or to use the libraries helping this stuff. 

On compatibility and spec-fitting, we are like covering all the avro spec without knowing how it works. We just put the record from `Avro.toRecord()` into the apache avro lib, and 🎉 it works. 
On tests, we are currently not in testing the avro format compliance, but the generated GenericRecords with another generic record. 

And the last axis, kotlin multi platform. Since a big part of the codebase and the mechanisms is highly linked to the apache avro Java lib, the multiplaform dream seems complicated to reach.

Now, how to improve that? Let kotlinx serialization do his best and why it has been created : easily encode whatever the kotlin object.
Have a look to the avro `Encoder` methods : https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/io/Encoder.java 
- wtiteInt
- writeBoolean
- writeString
- writeArray[Start/End] 
- writeMap[Start/End]
- etc...

All the methods fits exactly to kotlinx encoders, and the way of calling them fits perfectly with kotlinx workflow : 
- beginCollection -> writeArrayStart/writeMapStart + setItemCount
- endStructure -> writeArrayEnd/writeMapEnd
- encodeElement -> startItem
- all the encodeXX for primitives -> same in avro encoder

The other advantage is that there is absolutely no primitive autoboxing, also no more callbacks hell thanks to direct use of avro Encoder as output. 

@thake, any comment on this? You are in this project since a long time. This will be a big refactoring, while it could be game changer.

Plan for making the changes:
- finish the kotlin native feature PR that is simplifying the codebase and using the latest features of kotlinx serialization 
- adapt all tests to not compare generic records, but binary data
- modify encoders to use avro Encoder from configuration (changeable with Binary, json, and maybe others) 
[...]

PR #168, opened 2023-09-15
Generalize encoding/decoding tests

thake:
The current unit tests for encoding have been refactored as part of #160. Duplicate tests have been removed, and some bugs have been fixed on the way.


src/test/kotlin/com/github/avrokotlin/avro4k/endecode/ArrayEncoderTest.kt
```diff
@@ -0,0 +1,51 @@
...
+fun arrayEncodingTests(encoderToTest: EnDecoder): TestFactory {
+   return wordSpec {
+      "en-/decoder" should {
+         "generate GenericData.Array for an Array<Boolean>" {
+            @Serializable
+            data class ArrayBooleanTest(val a: Array<Boolean>)
+            val value = ArrayBooleanTest(arrayOf(true, false, true))
+            encoderToTest.testEncodeDecode(value, record(value.a.asList()))
+         }
+         "support GenericData.Array for an Array<Boolean> with other fields" {
```
Chuckame, line 22: For these tests that expect the same as the other but including other fields, maybe we could have a big data class with all the possible types (null/non-null, raw types, generics, enums, arrays, nested data classes, ...), it would naturally check that encoding and decoding are working even with complex objects and other fields. We would re-use the same as what I said just before (e.g. having a test method that handle both the data class instance and its avro representation).

Also, it would act as a full example of what can do the avro4k library.

Chuckame (changes requested):
Hello back, happy new year. Happy to come back on the project, I didn't have that much time until now! Feel free to tell me if you want me pushing a commit on this branch if you don't have time or di not understand :)

src/test/kotlin/com/github/avrokotlin/avro4k/endecode/EnDecoder.kt
```diff
@@ -0,0 +1,125 @@
...
+    fun avroLibDecoder(schema: Schema, inputStream: InputStream) : Decoder = 
+        DecoderFactory.get().jsonDecoder(schema, inputStream)
+}
+
+inline fun <reified T> EnDecoder.testEncodeDecode(
+    value: T,
+    shouldMatch: RecordBuilderForTest,
+    serializer: KSerializer<T> = avro.serializersModule.serializer<T>(),
+    schema: Schema = avro.schema(serializer)
+) {
```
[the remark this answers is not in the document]
thake: I had in mind that I took this problem into account, but I'm not sure anymore because life has gone on since I submitted this PR ;)

After a quick look at the code, I think this should not be a problem because the serialization result will always be compared against the `shouldMatch` parameter, which is the expected result as an Avro (Apache Avro) record. So, the encoded result will be compared against the reference implementation. I think this is the best we can do.
As an additional safeguard, we could introduce snapshot tests, but I had a bad experience with this kind of tests. They can become quite a hassle to properly maintain. 

I put it into a symmetric test clause because sometimes an Encoder/Decoder pair was not properly tested for each test case in both directions.

src/test/kotlin/com/github/avrokotlin/avro4k/endecode/ArrayEncoderTest.kt
thake: This is a good additional test case, but I would rather leave the specific test cases also in our test suite.

src/test/kotlin/com/github/avrokotlin/avro4k/endecode/EnDecoder.kt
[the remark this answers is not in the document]
Chuckame: Oh sorry, I did not got it during the first read.

I also feel a bit complicated the `EnDecoder` class, maybe a bit too complex for what it is in real.

Why not just having 3 methods `testEncodeDecode`, `testEncodeIsEqual` and `testDecodeIsEqual` without this `includeForEveryEncoder` and `EnDecoder` and `AvroLibEnDecoder` ? I would say that we don't want multiple implementations inside avro4k to not have a too complex project to maintain 😄 

If you are not taking this in the next days, I'll push a commit to show an idea 🚀

src/test/kotlin/com/github/avrokotlin/avro4k/endecode/EnDecoder.kt
```diff
@@ -0,0 +1,125 @@
...
+    val encoded = testEncodeIsEqual(value, shouldMatch, serializer, schema)
+    testDecodeIsEqual(encoded, value,serializer, schema)
+}
```
Chuckame, line 88: Here is what I'm talking about: taking the expected avro representation, encoding it with avro lib, and then use it to test our decoding part
```suggestion
inline fun <reified T> EnDecoder.testEncodeDecode(
    value: T,
    shouldMatch: RecordBuilderForTest,
    serializer: KSerializer<T> = avro.serializersModule.serializer<T>(),
    schema: Schema = avro.schema(serializer)
) {
    testEncodeIsEqual(value, shouldMatch, serializer, schema)
    testDecodeIsEqual(avroLibBinaryEncoding(shouldMatch.toAvro(schema)), value, serializer, schema)
}
```

Chuckame:
Let's merge it as it is, as I'm currently adding ktlint, spotless and reformat everything, and also moving forward on the v2, so having simpler tests will be awesome.

I'm not a fan of the open door of multiple EnDecode implementation as it complexifies for a need that we don't have now, and mostly avro4k should not have multiple way of doing avro or users won't really know what to chose.

If we want to test decoding and encoding in multiple container types (single object, confluent schema registry, ...), let's tackle some "wrapping" api to be able of decorate a binary avro message with whatever we need.

I'll do a little other PR to fix this point after all the code reformat

PR #171, opened 2023-11-17
Added support for decoding with avro alias

### LINKED ISSUE #170: RecordDecoder including aliases from AvroAlias annotation for possible field names
Problem: We want to read some data using Avro evolution mechanism for field renaming. Current behavior is to use the property name, serial name, or avro name only which does not account for AvroAlias annotation.

Proposed solution: fieldValue() function iterates through list of possible field names, starting with the resolvedFieldName. For each attempt, if non-null value returned from record.get(...) then return it, else continue.

trdw:
Problem: We want to read some data using Avro evolution mechanism for field renaming. Current behavior does not account for AvroAlias annotation.

Proposed solution: fieldValue() function iterates through list of possible field names, starting with the resolvedFieldName. For each attempt, if record contains the field name, return the value returned from record.get(...), else continue.

Closes #170

### DIFF
--- a/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNaming.kt
+++ b/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNaming.kt
@@ -21,4 +21,11 @@ class FieldNaming(private val name: String, annotations: List<Annotation>) {
     *  Takes into account @AvroName.
     */
    fun name(): String = extractor.name() ?: name
+
+
+   /**
+    *  Returns the avro aliases for the current element.
+    *  Takes into account @AvroAlias.
+    */
+   fun aliases(): List<String> = extractor.aliases()
 }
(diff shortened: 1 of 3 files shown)


Chuckame:
Hello, sorry for the delay, I'll take a lot at it to see if it is not changing some other behavior. Thanks for the PR 🚀

Chuckame:
AFAIK, without your PR, we are able to decode an old schema written we another alias while the kotlin data class uses the real field name (not alias).

What is not is when you read a new schema where the kotlin data class is using an alias for its field name instead of the real field name. That's because apache avro library "normalize" the field name, and whatever the alias name or the real name, it creates a GenericRecord using field's real name (that is totally normal).

I don't see any particular issue, I approve it. @thake it's ok for you also ?

Chuckame (approved):

Chuckame:
Side note: We should cache somewhere with a map like `Map<alias, real name>` to prevent useless aliases iterations. Let's plan it for `v2` @thake

PR #173, opened 2024-01-21
chore(build): Replace buildSrc by gradle's versionCatalogs

Chuckame:
Removed the custom buildSrc in favor of gradle's `versionCatalogs`.

I also removed sksamuel as he is not a maintainer of this avro4k library.


build.gradle.kts
```diff
@@ -122,16 +123,16 @@ publishing {
             }
 
             developers {
-               developer {
```
thake, line 125: I would keep sksamuel here because most of the avro4k code originates from him.

build.gradle.kts
```diff
@@ -122,16 +123,16 @@ publishing {
...
-                  id.set("sksamuel")
-                  name.set("Stephen Samuel")
-                  email.set("sam@sksamuel.com")
-               }
                developer {
                   id.set("thake")
                   name.set("Thorsten Hake")
                   email.set("mail@thorsten-hake.com")
                }
+               developer {
```
thake, line 131: Very good :)

thake (changes requested):
Thanks for the PR. I was thinking about this change but never had the time to go for it.

build.gradle.kts
Chuckame: The developers are the maintainers, why putting him there? We still put the credits on the Readme, and for sure the code will evolve a lot, that will reduce more and more the original code 🤔

Chuckame:
Just a side note, I would prefer that you approve instead of merging yourself my PR 😄 I usually read again what I've did before merging to be sure that I missed nothing !

Issue #141, opened 2023-05-23
Support avro single object encoding


jangalinski:
Currently, the README states that single object encoding is not supported.

Following up on the discussion in #138, I believe this should be possible to implement.

From the [Specification](https://avro.apache.org/docs/1.11.1/specification/#single-object-encoding-specification):

Single Avro objects are encoded as follows:

> 1. A two-byte marker, C3 01, to show that the message is Avro and uses this single-record format (version 1).
> 1. The 8-byte little-endian CRC-64-AVRO [fingerprint](https://avro.apache.org/docs/1.11.1/specification/#schema-fingerprints) of the object’s schema.
> 1. The Avro object encoded using [Avro’s binary encoding](https://avro.apache.org/docs/1.11.1/specification/#binary-encoding).

Binary encoding is already supported, and the fingerprint of the schema should be accessible from 

`SchemaNormalization.parsingFingerprint64(Avro.default.schema(MySerializableDataClass.serializer()))`

so serialization should come down to concatenating  bytes.

For deserialization, we need to deconstruct the single object encoded byte array, check the first two marker bytes, extract the fingerprint, lookup the writer schema from a `SchemaStore` and decode the remaining binary payload to GenericRecord and then to the target data class.

jangalinski:
I will try to provide this feature.

jangalinski:
see #142 142

Chuckame:
Hello, after some work at revamping the `Avro` api #186 (WIP), there is a new easier and promising way of implementing the single object (for the moment, named serialization delegate). The avro input and output stream would be removed. Don't hesitate to comment it as you worked hard to integrate this single object format in the current version of avro4k.

Thanks for being patient and sorry for the delay again.

PR #186, opened 2024-02-07
feat: New Avro entrypoint

### LINKED ISSUE #163: Beautify the Avro api entrypoint
Currently, to do common stuff, we have to start by `Avro.default`. But when we want to override defaults, we do like this:
```kotlin
Avro(SerializersModule {
    contextual(...)
}, AvroConfiguration(
    namingStrategy = ...
))
```

kolinx serialization Json (and other official supported formats) are doing very interesting stuff, here the examples:
```kotlin
// default way
Json.encodeToByteArray(...)

// customized way
Json {
   serializersModule = SerializersModule {
    contextual(...)
   }
    namingStrategy = ...
}.encodeToByteArray(...)
```

It would be great to do it ! It's not that complex to implement, it's mainly copy/paste from [Json class](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt)

Chuckame:
Here is the new proposed avro entrypoint 🥳

### Goal
Provide an unified, easy and extensible way of encoding and decoding avro. Unified and easy as it should only have methods on the `Avro` class (no streams, no builders, ...). Extensible as avro can be used in multiple contexts, and some requires adding some metadata (data files) or schema info (confluent schema registry, single object).

Closes #163 #141

### Requirements to merge:
- Encode everything at root (not only records)
- Encode natively to bytes

### Breaking changes:
- No more `AvroInputStream` and `AvroOutputStream`, thanks to the `encodeSequenceXxx` and `decodeSequenceXxx` replacements with `AvroSerializationDelegate`
- No more encoding to and from Generic data, as the final goal of this library is to encode and decode in avro, not in Generic data, and also another step of removing the java avro library (no more public api in java, except for the Schemas)
- `encodeToByteArray` and `decodeFromByteArray` will now be pure avro (no header, just the encoded data itself without decorators) while before it was in `Data` mode (avro data file format) 

### Tickets to open after that:
- Create the necessary serializers to encode and decode Generic data (from bytes to generic, and from generic to bytes)

### The methods
Each base method having a `Schema` and a `SerializerStrategy` or `DeserializerStrategy` has its related extension method to automagically get the `SerializerStrategy` or `DeserializerStrategy` and the schema from the reified type.

We now fully support java streams and ByteArray.

We are now able of streaming the avro contents with the new `encodeSequence` and `decodeSequence` methods, useful for reading avro files by example!

image
image
image

### Example of usage
- Just encode a string payload
```kotlin
val bytes = Avro.encodeToByteArray("a value to be encoded")
val value = Avro.decodeFromByteArray<String>(bytes)
```
- Just encode a record payload
```kotlin
val bytes = Avro.encodeToByteArray(ARecord(stringField = "something", intField = 42))
val value = Avro.decodeFromByteArray<ARecord>(bytes)
```
- Encode to the avro data file format
```kotlin
val avro = Avro {
    serializationDelegate = Builtins.DataFile
}
File("file.avro").outputStream().use { outputStream ->
    AvroObjectContainerFile().encodeToStream(sequenceOf("a", "b", "c"), outputStream)
}
File("file.avro").inputStream().use { inputStream ->
    val sequence = avro.decodeFromStream<String>(inputStream)
}
```
- Just generate a schema from a type:
```kotlin
val schema = Avro.schema<MySerializedType>()
```


jangalinski:
Seems like you are aiming for static access, while in the current version, there is a default instance (Avro.default) to use, but also the option to create a custom instance. This is quite important when working with custom logical types .. am I missing st. or would we loose the option of configuring a custom instance?

PR #247, opened 2024-08-09
Add migration from AvroEncodeFormat.Binary example

### LINKED ISSUE #245: How to read previously written Binary files from v1
In previous version there was a AvroEncodeFormat.Binary. Using that you could store data without schema and it was possible to provide an external schema on read. 
I tried to read previously saved data using AvroSingleObject and the provided schema, but the magic numbers do not match, so it won't work.
Have I missed something because I can't find a way to continue supporting previously saved data using the new version of avro4k. 
Thank you for your great work!

avro4k 1.10.1:
```
    /**
     * Encodes the record in a binary format without schema information, the most compact format.
     *
     * See https://avro.apache.org/docs/current/spec.html#binary_encoding
     */
    object Binary : AvroEncodeFormat() {
        override fun <T> createOutputStream(
            output: OutputStream,
            schema: Schema,
            converter: (T) -> GenericRecord
        ) = AvroBinaryOutputStream(output, converter, schema)

    }
```

rutkowskij:
Closes #245

### DIFF
--- a/Migrating-from-v1.md
+++ b/Migrating-from-v1.md
@@ -166,6 +166,21 @@ Files.newOutputStream(Path("/your/file.avro")).use { outputStream ->
 }
 ```
 
+## Reading previously written binary encoded (`AvroEncodeFormat.Binary`) files
+
+It was possible to use the `AvroEncodeFormat.Binary` format, which used [binary encoding](https://avro.apache.org/docs/current/spec.html#binary_encoding). The data did not have an embedded schema, so it had to be specified when reading.
+
+```kotlin
+// Previously
+Avro.default.openInputStream(serializer) { decodeFormat = AvroDecodeFormat.Binary(schema) }
+    .from(data).use { avroInputStream -> return avroInputStream.nextOrThrow() }
+
+// Now
+val reader = GenericDatumReader<Any>(schema)
+val decoder = DecoderFactory.get().binaryDecoder(ByteArrayInputStream(data), null)
+Avro.decodeFromGenericData(schema, serializer, reader.read(null, decoder))
+```
+
 ## Writing a collection of Byte as BYTES or FIXED
 
 If you really want to encode a `BYTES` type, just use the `ByteArray` type or write your own `AvroSerializer` to control the schema and its serialization.
--- a/Migrating-from-v1.md
+++ b/Migrating-from-v1.md
@@ -176,9 +176,17 @@ Avro.default.openInputStream(serializer) { decodeFormat = AvroDecodeFormat.Binar
     .from(data).use { avroInputStream -> return avroInputStream.nextOrThrow() }
 
 // Now
-val reader = GenericDatumReader<Any>(schema)
-val decoder = DecoderFactory.get().binaryDecoder(ByteArrayInputStream(data), null)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -131,7 +131,7 @@ fun main() {
 
     // Deserializing objects
     Files.newInputStream(Path("your-file.bin")).use { fileStream ->
-        AvroObjectContainer.decodeFromStream<Project>(valuesToEncode, fileStream, builder).forEach {
+        AvroObjectContainer.decodeFromStream<Project>(fileStream).forEach {
```
Chuckame, line 134: I appreciate documentation improvements, many thanks 👏

Migrating-from-v1.md
```diff
@@ -166,6 +166,21 @@ Files.newOutputStream(Path("/your/file.avro")).use { outputStream ->
...
+```kotlin
+// Previously
+Avro.default.openInputStream(serializer) { decodeFormat = AvroDecodeFormat.Binary(schema) }
+    .from(data).use { avroInputStream -> return avroInputStream.nextOrThrow() }
+
+// Now
+val reader = GenericDatumReader<Any>(schema)
+val decoder = DecoderFactory.get().binaryDecoder(ByteArrayInputStream(data), null)
+Avro.decodeFromGenericData(schema, serializer, reader.read(null, decoder))
+```
```
Chuckame, line 182: I think you are mixing formats and the avro4k api.

The previous `Binary` format is now handled by `Avro` as it is the pure avro serialization. To encode many elements in "binary", now you need to call as many element as you have the `Avro.encodeToStream` method. To decode, it's the same: loop on `Avro.decodeFromStream` until there is no remaining bytes in the input stream. Side note: you should not encode many times using `Avro` in the same file/stream, as it's unstructured content which needs maintenance, version control, and mainly the schema! The `AvroObjectContainer` spec has been created exactly for that purpose.

The previous `Data` format is now handled by `AvroObjectContainer` which is the official name in the specs.

Btw, all the avro serializations... Are binary, that's why this concept has been removed in the v2.

Migrating-from-v1.md
```diff
@@ -166,6 +166,21 @@ Files.newOutputStream(Path("/your/file.avro")).use { outputStream ->
...
+
+```kotlin
+// Previously
+Avro.default.openInputStream(serializer) { decodeFormat = AvroDecodeFormat.Binary(schema) }
+    .from(data).use { avroInputStream -> return avroInputStream.nextOrThrow() }
+
+// Now
+val reader = GenericDatumReader<Any>(schema)
+val decoder = DecoderFactory.get().binaryDecoder(ByteArrayInputStream(data), null)
+Avro.decodeFromGenericData(schema, serializer, reader.read(null, decoder))
```
Chuckame, line 181:
```suggestion
val inputStream = ByteArrayInputStream(data)
while (inputStream.remaining() > 0) {
    // If the writer schema corresponds to the specified type
    val element = Avro.decodeFromStream<MyType>(inputStream)

    // If the writer schema does not correspond to the specified type
    val element = Avro.decodeFromStream<MyType>(writerSchema, inputStream)

    // With explicit writer schema and serializer
    val element = Avro.decodeFromStream(writerSchema, serializer, inputStream)
}

Chuckame (approved):

PR #270, opened 2025-01-31
Lists of ints serialization benchmarks

rutkowskij:
### Observed Performance Degradation During Deserialization of Objects with Large Collections
I have observed a significant performance degradation when deserializing objects that contain large collections (lists). In my case, the deserialization time increased by nearly `3x` 🧐
Interestingly, when deserializing previously created generic objects, the performance drop is not as noticeable. The most significant difference compared to the previous version is for objects serialized with version `2.1.1` and deserialized with version 2.1.1.
I added benchmarks on my local branch to measure the performance in version `1.10.1` and compared it to the current implementation (entries with name `Avro4kV1*`). I created a very simple model that contained just an Int list, and the performance drop was already noticeable.
I have added the results to the README for comparison. The remaining results (for v2) are comparable to those in the README on my machine, so I think I can include the results for reference.
Interestingly, there was a strange and intriguing result for JacksonAvro serialization (the fastests one). I’m not sure how to interpret it at this moment, but this effect does not seem to occur with more complex models than just an Int list.


benchmark/README.md
```diff
@@ -40,11 +40,26 @@ c.g.a.b.simple.Avro4kSimpleBenchmark.read                      thrpt    5  17406
...
 c.g.a.b.simple.Avro4kGenericWithApacheAvroSimpleBenchmark.write thrpt   5  190365.959 ±  1014.944  ops/s  -58.60%
 c.g.a.b.simple.JacksonAvroSimpleBenchmark.write                thrpt    5  131492.564 ± 10888.843  ops/s  -71.40%
+c.g.a.b.simple.Avro4kV1SimpleBenchmark.write                   thrpt    5   54429.114 ±   963.765  ops/s  -88.16%
+
+
+c.g.a.b.lists.Avro4kV1ListsBenchmark.read	                   thrpt	5	  475.048 ±    22.037  ops/s	   0%
+c.g.a.b.lists.ApacheAvroReflectListsBenchmark.read	           thrpt	5	  290.714 ±     6.295  ops/s  -38.80%
+c.g.a.b.lists.Avro4kGenericWithApacheAvroListsBenchmark.read   thrpt	5	  311.651 ±     3.760  ops/s  -34.42%
+c.g.a.b.lists.Avro4kListsBenchmark.read	                       thrpt    5     412.503 ±     6.127  ops/s  -13.16%
+c.g.a.b.lists.JacksonAvroListsBenchmark.read	               thrpt    5    1002.527 ±    23.039  ops/s +111.09%
```
Chuckame, line 56: Here are the results on my computer using the new kotlinx-io decoder, doubling the performances!! Still not better than jackson, but it's close!

```suggestion
c.g.a.b.lists.Avro4kListsBenchmark.read	                       thrpt    5     813.448 ±     19.265  ops/s
c.g.a.b.lists.JacksonAvroListsBenchmark.read	               thrpt    5    972.243 ±    59.303  ops/s
```

Chuckame:
can you rebase on main, which is now including the cleaning of benchmarks, and also the little optimisation of non-direct encoder/decoder ?

jakubrutkowski-tomtom:
> can you rebase on main, which is now including the cleaning of benchmarks, and also the little optimisation of non-direct encoder/decoder ?
> 
> That way I'll be able to work improving performances without asking to you to update the results at each change 😅

done

rutkowskij:
> can you rebase on main, which is now including the cleaning of benchmarks, and also the little optimisation of non-direct encoder/decoder ?
> 
> That way I'll be able to work improving performances without asking to you to update the results at each change 😅

done

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/ManualProfiling.kt
```diff
@@ -26,4 +29,17 @@ internal object ManualProfilingRead {
...
+    @JvmStatic
+    fun main(vararg args: String) {
+        Avro4kListsBenchmark().apply {
+            initTestData()
+            for (i in 0 until 1_000) {
+                if (i % 100 == 0) println("Iteration $i")
+                read()
+            }
+        }
+    }
```
Chuckame, line 44: Can you remove this part ? This file, even if it's not that explicit, is to run a particular benchmark and identify bottlenecks using a profiler. It's hacky for sure, I'll refactor it later to allow executing any benchmark separately

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/internal/ListWrapperDataClass.kt
```diff
@@ -0,0 +1,33 @@
...
+@Serializable
+internal data class StatsEntry(val elementId: ElementId, val values: List<StatValue>) {
+    companion object {
+        fun create(entriesSize: Int, random: RandomUtils = RandomUtils()) =
+            StatsEntry(random.nextLong(), List(entriesSize) { random.nextLong() })
+    }
+}
+
+typealias ElementId = Long
+typealias StatValue = Long
```
Chuckame, line 22: use internal everywhere to not change the apiDump

```suggestion
internal typealias ElementId = Long
internal typealias StatValue = Long
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/ApacheAvroReflectListsBenchmark.kt
```diff
@@ -0,0 +1,47 @@
...
+    @Benchmark
+    fun read() {
+        if (writeMode) writeMode = false
+        val decoder = DecoderFactory.get().directBinaryDecoder(ByteArrayInputStream(data), null)
+        reader.read(null, decoder)
+    }
+
+    @Benchmark
+    fun write() {
+        if (!writeMode) writeMode = true
```
Chuckame, line 44:
```suggestion
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/ApacheAvroReflectListsBenchmark.kt
```diff
@@ -0,0 +1,47 @@
...
+        reader = ReflectData.get().createDatumReader(schema) as DatumReader<ListWrapperDatasClass>
+    }
+
+    override fun prepareBinaryData() {
+        data = Avro.encodeToByteArray(schema, lists)
+    }
+
+    @Benchmark
+    fun read() {
+        if (writeMode) writeMode = false
```
Chuckame, line 37:
```suggestion
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/ApacheAvroReflectListsBenchmark.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.ByteArrayInputStream
+import java.io.OutputStream
+
+internal class ApacheAvroReflectListsBenchmark : SerializationListsBenchmark() {
+    lateinit var writer: DatumWriter<ListWrapperDatasClass>
+    lateinit var encoder: Encoder
+    lateinit var reader: DatumReader<ListWrapperDatasClass>
+
+    lateinit var data: ByteArray
+    var writeMode = false
```
Chuckame, line 22:
```suggestion
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/ApacheAvroReflectListsBenchmark.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    override fun prepareBinaryData() {
+        data = Avro.encodeToByteArray(schema, lists)
+    }
+
+    @Benchmark
+    fun read() {
+        if (writeMode) writeMode = false
+        val decoder = DecoderFactory.get().directBinaryDecoder(ByteArrayInputStream(data), null)
```
Chuckame, line 38:
```suggestion
        val decoder = DecoderFactory.get().binaryDecoder(data, null)
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/Avro4kGenericWithApacheAvroListsBenchmark.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override fun prepareBinaryData() {
+        data = Avro.encodeToByteArray(schema, lists)
+    }
+
+    @OptIn(ExperimentalSerializationApi::class)
+    @Benchmark
+    fun read() {
+        if (writeMode) writeMode = false
+        val decoder = DecoderFactory.get().directBinaryDecoder(ByteArrayInputStream(data), null)
```
Chuckame, line 42:
```suggestion
        val decoder = DecoderFactory.get().binaryDecoder(data, null)
```

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/Avro4kGenericWithApacheAvroListsBenchmark.kt
```diff
@@ -0,0 +1,54 @@
...
+import java.io.ByteArrayInputStream
+import java.io.OutputStream
+
+internal class Avro4kGenericWithApacheAvroListsBenchmark : SerializationListsBenchmark() {
+    lateinit var writer: DatumWriter<Any?>
+    lateinit var encoder: Encoder
+    lateinit var reader: DatumReader<Any?>
+
+    lateinit var data: ByteArray
+    var writeMode = false
```
Chuckame, line 25: You can remove the `writeMode` everywhere, I don't know why it was there in the other benchmarks 😅

Chuckame (changes requested):

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/ManualProfiling.kt
rutkowskij: That's exactly how I used it :D Removing...

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/internal/ListWrapperDataClass.kt
rutkowskij: I added these type aliases here to make sure that it doesn't affect anything because I have a similar problem in the code where I originally noticed the problem. I wonder if I should leave it as is or simplify it (eg first version with just list of ints 
 which I think shows the issue without any additional complexity needed)?

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/lists/ApacheAvroReflectListsBenchmark.kt
rutkowskij: I forgot to update list benchmark to align with latest changes - fixing all

benchmark/src/main/kotlin/com/github/avrokotlin/benchmark/internal/ListWrapperDataClass.kt
Chuckame: No keep both, but you can remove the typealias, finally the main issue is the low-level encoding/decoding of ints. I'll also rename your issue and PR

benchmark/README.md
```diff
@@ -44,6 +44,19 @@ c.g.a.b.simple.Avro4kSimpleBenchmark.write                      thrpt    5  4466
 c.g.a.b.simple.ApacheAvroReflectSimpleBenchmark.write           thrpt    5  320367.673 ± 33394.537  ops/s
 c.g.a.b.simple.Avro4kGenericWithApacheAvroSimpleBenchmark.write thrpt    5  168702.542 ±  5553.797  ops/s
 c.g.a.b.simple.JacksonAvroSimpleBenchmark.write                 thrpt    5  138898.312 ±  9156.715  ops/s
+
+c.g.a.b.lists.Avro4kV1ListsBenchmark.read	                    thrpt	 5	   475.048 ±    22.037  ops/s	    0%
+c.g.a.b.lists.ApacheAvroReflectListsBenchmark.read	            thrpt	 5	   290.714 ±     6.295  ops/s  -38.80%
+c.g.a.b.lists.Avro4kGenericWithApacheAvroListsBenchmark.read    thrpt	 5	   311.651 ±     3.760  ops/s  -34.42%
+c.g.a.b.lists.Avro4kListsBenchmark.read	                        thrpt    5     412.503 ±     6.127  ops/s  -13.16%
+c.g.a.b.lists.JacksonAvroListsBenchmark.read	                thrpt    5    1002.527 ±    23.039  ops/s +111.09%
```
Chuckame, line 52: Is this result taking the last version on `main` ? 😮 😭

Chuckame (approved):

benchmark/README.md
jakubrutkowski-tomtom: Unfortunately not. It shows results 1.10.1 vs 2.1.1 released versions

rutkowskij: Nope. It shows results for released versions - v1.10.1 vs 2.1.1

Chuckame: ok i'll re-run to compare using the main version

PR #332, opened 2025-07-18
docs: README doc fixes

felixwiemuth:


README.md
```diff
@@ -77,8 +77,9 @@ data class Project(val name: String, val language: String)
 
 fun main() {
     val schema = Avro.schema<Project>()
-    val schemasByFingerprint = mapOf(SchemaNormalization.parsingFingerprint64(schema), schema)
-    val singleObjectInstance = AvroSingleObject { schemasByFingerprint[it] }
+    val schemasByFingerprint = mapOf(SchemaNormalization.parsingFingerprint64(schema) to schema)
+    @OptIn(ExperimentalSerializationApi::class)
```
Chuckame, line 81: This does not add anything interesting, and can be removed by the user in his preferred way. Could you remove it?
```suggestion
```

Chuckame (changes requested):
Thanks for fixing typos! Just a little feedback

README.md
felixwiemuth: I think (don't remember exactly) that it is a compilation error if you don't opt-in - therefore, and for the sake of conveying that this is experimental, I thought it is a good idea of having it included in the example?

Feel free to commit the removal or not :)

Issue #317, opened 2025-07-01
Allow encoding/decoding RECORD as generic Map


Chuckame:
**Is your feature request related to a problem? Please describe.**
When the codebase doesn't have the corresponding or compatible data class, the user isn't able to decode or encode generic avro data. At that specific moment, the only way to do generic decoding is by using apache's avro `GenericData` decoding. Same for encoding.

**Describe the solution you'd like**
The idea is to allow encoding and decoding `Map` as RECORD.

Encoding a `Map` is quite easy to do, as it would reuse the same workflow as RECORD encoding, except for the UNION, where we need to match the closest RECORD regarding the field names. Maybe as a first step, taking the first RECORD in the UNION matching at least the number of fields would be enough.

Decoding is much more tricky, as we have to take into account the logical types and the defaults. It would allow decoding `Any`.

**Additional context**
This would make another step reducing the need of the `GenericData` apache's api, also enabling easiest genericity for kafka-serializer #218

PR #337, opened 2025-08-04
feat: Allow generic de-serializing of Any type

### LINKED ISSUE #336: Decode (and encode) Any
Currently, it's not possible to decode generically, without knowing the decoded type or the DeserializationStrategy at compile time. Same for encoding, where it needs to know the type or the SerializationStrategy at compile time too. 

As a workaround, we can use the standard apache library to decode into their generic classes, like GenericRecord, GenericEnumSymbol, GenericFixed, but that doesn't enable all the kotlin classes support, and requires twice the knowledge : avro4k and apache avro java library.

But to fulfill the goal of becoming a multiplatform kotlin library #207, and enable a unique entrypoint to any avro (de)serialization, we need to allow deserializingto `Any` type. This also allows doing #218, as it needs to encode and decode to a type known only at runtime. 

By decoding to `Any` type, we can read any avro type, without complex configuration. Also, we don't want to mess up the corresponding kotlin types:
- scalar avro types `null`, `boolean`, `int`, `long`, `float`, `double`, `string`, `bytes` becomes kotlin primitives, respectively `null`, `Boolean`, `Int`, `Long`, `Float`, `Double`, `String`, `ByteArray`
- `array` becomes `ArrayList`
- `map` becomes `HashMap`
- `fixed` becomes `ByteArray`
- `enum` symbol becomes `String`
- `record` becomes a `HashMap` where keys are the record's fields name

At encoding, it also allows to encode as `Any` type when the value type is known only at runtime.

For both encoding and decoding, avro4k will require the writer schema to be always explicitly set, or it won't be able to infer it.

Chuckame:
Closes #336 #317

### DIFF
--- a/api/avro4k-core.api
+++ b/api/avro4k-core.api
@@ -35,24 +35,27 @@ public final class com/github/avrokotlin/avro4k/AvroBuilder {
 	public final fun setFieldNamingStrategy (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;)V
 	public final fun setImplicitEmptyCollections (Z)V
 	public final fun setImplicitNulls (Z)V
+	public final fun setLogicalTypeSerializer (Ljava/lang/String;Lkotlinx/serialization/KSerializer;)V
 	public final fun setSerializersModule (Lkotlinx/serialization/modules/SerializersModule;)V
 	public final fun setValidateSerialization (Z)V
 }
 
 public final class com/github/avrokotlin/avro4k/AvroConfiguration {
 	public fun <init> ()V
-	public fun <init> (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;ZZZ)V
-	public synthetic fun <init> (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;ZZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;ZZZLjava/util/Map;)V
+	public synthetic fun <init> (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;ZZZLjava/util/Map;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;
 	public final fun component2 ()Z
 	public final fun component3 ()Z
 	public final fun component4 ()Z
-	public final fun copy (Lcom/github/avrokotlin/avro4k/FieldNamingStrategy;ZZZ)Lcom/github/avrokotlin/avro4k/AvroConfiguration;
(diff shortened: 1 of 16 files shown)


api/avro4k-core.api
```diff
@@ -310,6 +313,23 @@ public abstract interface class com/github/avrokotlin/avro4k/ShortValueDecoder {
...
 
+public class com/github/avrokotlin/avro4k/serializer/AnySerializer : kotlinx/serialization/KSerializer {
+	public static final field Default Lcom/github/avrokotlin/avro4k/serializer/AnySerializer$Default;
+	public fun <init> ()V
+	public fun deserialize (Lkotlinx/serialization/encoding/Decoder;)Ljava/lang/Object;
+	public fun getDescriptor ()Lkotlinx/serialization/descriptors/SerialDescriptor;
+	public final fun getNullable ()Lkotlinx/serialization/KSerializer;
+	protected fun inferSerializationStrategyFromNonSerializableType (Lkotlinx/serialization/modules/SerializersModule;Ljava/lang/Class;)Lkotlinx/serialization/SerializationStrategy;
+	protected fun preResolveDeserializationStrategy (Lkotlinx/serialization/modules/SerializersModule;Lorg/apache/avro/Schema;)Lkotlinx/serialization/DeserializationStrategy;
+	protected fun resolveEnumDeserializationStrategy (Lkotlinx/serialization/modules/SerializersModule;Lorg/apache/avro/Schema;)Lkotlinx/serialization/DeserializationStrategy;
```
Chuckame, line 324: Annotate with internal

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/AnySerializer.kt
```diff
@@ -0,0 +1,216 @@
...
+        if (Map::class.java.isAssignableFrom(type)) {
+            return MapSerializer(this@AnySerializer.nullable, this@AnySerializer.nullable) as KSerializer<Any>
+        }
+
+        return inferSerializationStrategyFromNonSerializableType(type) as SerializationStrategy<Any>?
+            ?: throw SerializationException("Cannot find serializer for type $type. Please register a serializer for this type in the serializers module.")
+    }
+
+    /**
+     * **During encoding**, called when avro4k did not find the corresponding [type]'s serializer in the [com.github.avrokotlin.avro4k.Avro.serializersModule].
```
Chuckame, line 98:
```suggestion
     * Called when avro4k did not find the corresponding [type]'s serializer in the [com.github.avrokotlin.avro4k.Avro.serializersModule].
```

src/main/kotlin/com/github/avrokotlin/avro4k/serializer/AnySerializer.kt
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * Generally useful when you don't know the exact type at runtime, but the interface.
+     *
+     * Results are cached, so expect only one call per [type].
+     *
+     * @return the serializer to use for the given [type], or `null` if none, which will raise a [SerializationException].
+     */
+    protected open fun SerializersModule.inferSerializationStrategyFromNonSerializableType(type: Class<out Any>): SerializationStrategy<*>? {
```
Chuckame, line 106: Add internal annotation to all open methods

PR #397, opened 2025-10-26
fix(gradle-plugin): Make the source generation compatible with the gradle config cache

khouari1:
Changes to gradle plugin to not reference `Project` type in the task execution logic. I have tested myself.

Thanks for developing a very useful plugin by the way!

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/github/avrokotlin/avro4k/plugin/gradle/Avro4kGradlePlugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/github/avrokotlin/avro4k/plugin/gradle/Avro4kGradlePlugin.kt
@@ -122,13 +122,13 @@ public abstract class GenerateKotlinAvroSourcesTask : DefaultTask() {
         }
 
         files.associateWith { file ->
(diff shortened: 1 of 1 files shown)


Chuckame:
You did not test using the provided pr but main, which was expected to still the issue. Anyway, in the mean time, I successfully tested it locally, reproduced it.

I released the 2.7.0 which should now be cacheable! Thanks for pointing it! I let you close the PR, or comment if you still have the issue on your side. 🙏

Chuckame:
After reading the requirements for build-cache, it seems that the only issue here is using the `project` instance as you already spotted. However, it's only for logging purpose, so I prefer totally removing the use of it, and do not make relative paths.

Chuckame:
Can you confirm @khouari1 that it works on your side ? Sadly on my, it still fail on another step:
```
1 problem was found storing the configuration cache.
- Task `:compileKotlin` of type `org.jetbrains.kotlin.gradle.tasks.KotlinCompile`: error writing value of type 'java.util.Collections$SingletonList'
```

Chuckame:
Hello @khouari1 , any news?

khouari1:
@Chuckame Sorry for the delay, I will check on this

PR #426, opened 2025-12-13
feat(plugin): Support mapping kotlin properties names to camelCase in generated code

### LINKED ISSUE #425: Add Field Naming Strategy Support to Gradle Plugin
## Is your feature request related to a problem? Please describe.
Currently, when generating Kotlin code using the avro4k Gradle plugin, Avro schema field names are used as-is for Kotlin property names.

## Describe the solution you'd like
I found `FieldNamingStrategy` in the project, and I’d like to add a `fieldNamingStrategy` option to the Gradle plugin to support field name transformation during code generation:

```kotlin
avro4k {
    sourcesGeneration {
        fieldNamingStrategy = "SnakeCase" // or "PascalCase", "OriginalElementName"
    }
}
```

Alternatively, we can allow a custom strategy like this:

```kotlin
object CustomStrategy : FieldNamingStrategy {
    override fun resolve(descriptor: SerialDescriptor, elementIndex: Int): String =
        "x_" + descriptor.getElementName(elementIndex)
}

avro4k {
    sourcesGeneration {
        inputSchemas.from(file("src/main/avro"))
        outputDir = file("build/generated/sources/avro/main")
        fieldNamingStrategyInstance.set(CustomStrategy)
    }
}
```

## Describe alternatives you've considered
Writing post-processing scripts to modify generated code

However, these approache is cumbersome and error-prone.

## Additional context
It would be great to provide a default NamingStrategy that is compatible with the one used by `gradle-avro-plugin`.

Lee-WonJun:
## Summary

This PR adds **useKotlinConventionForFieldNames** to the Avro4k Gradle plugin for mapping avro record names compliant with the kotlin standards.
Previously, Avro schema field names were used verbatim when generating Kotlin property names.
With this change, users can transform field names during code generation to better align with Kotlin naming conventions.

## Changes

* Introduced a new `NameStrategy` interface for code generation

  * Built-in strategies: `IDENTITY`, `CAMEL_CASE`
  * Support for custom strategies
* Since Gradle cannot detect changes in custom strategies,
  a name strategy identifier was added to **force invalidation** of the incremental build cache (I couldn’t find a better approach for handling interface-based configuration in the plugin)
* Updated `KotlinGenerator` to apply the configured naming strategy
* Default behavior remains unchanged (`IDENTITY`)

## Note

Although an existing `FieldNamingStrategy` exists, it does not appear to be intended for Avro → Kotlin code generation.
Because code generation has different requirements, this PR introduces a simpler and more deterministic `FieldNamingStrategy` tailored specifically for this use case.

## Related Issue
Closes #425


kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/NameStrategy.kt
```diff
@@ -0,0 +1,50 @@
...
+    public companion object {
+        public val IDENTITY: NameStrategy = NamedNameStrategy("identity") { it }
+        public val CAMEL_CASE: NameStrategy = NamedNameStrategy("camelCase") { it.toCamelCase() }
+        public val SNAKE_CASE: NameStrategy = NamedNameStrategy("snakeCase") { it.toSnakeCase() }
+        public val PASCAL_CASE: NameStrategy = NamedNameStrategy("pascalCase") { it.toPascalCase() }
+
+        public fun custom(identifier: String, formatter: (String) -> String): NameStrategy =
+            NamedNameStrategy(identifier, formatter)
+    }
+}
```
Chuckame: Let's not overthink it. Custom naming strategy complexifies the whole implementation, as you need to pass both the naming strategy name (or its identifier in your implementation), a serializable object, and maybe the formatter if custom. Instead, you should prefer having an enum on the plugin side, mapping to the corresponding built-in naming strategy using 
```
when (namingStrategyProperty) {
   BuiltinNamingStrategy.IDENTITY -> NamingStrategy.Identity
   BuiltinNamingStrategy.CAMEL_CASE -> NamingStrategy.CamelCase
   // ...
}
```

If the actual avro schemas names are too weird to be used in kotlin codebase, I would really more enforce having a mapping layer to not deviate too much between the (avro) serialization layer and inside the app.

Also, I would prefer using interface + specific object extending it instead of a lambda. If the user needs custom strategy, he then implement the interface on his side.

Finally, just a detail, but let's name it a `NamingStrategy`, and use objects for each implementation.


```suggestion
public fun interface NamingStrategy {
    public fun format(original: String): String

    public companion object {
        public object Identity : NamingStrategy { ... }
        public object CamelCase : NamingStrategy { ... }
        public object SnakeCase : NamingStrategy { ... }
        public object PascalCase : NamingStrategy { ... }
    }
}
```

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/NameStrategy.kt
```diff
@@ -0,0 +1,50 @@
...
+internal fun String.toCamelCase(): String = toPascalCase().lowercaseFirstChar()
+
+internal fun String.toSnakeCase(): String {
+    return trim()
+        .replace(Regex("([a-z0-9])([A-Z])"), "$1_$2")
+        .replace(Regex("[\\W]+"), "_")
+        .replace(Regex("__+"), "_")
+        .trim('_')
+        .lowercase()
+}
```
Chuckame: Some remarks:
- prefer using only one regex, and store it in a `val` outside the method to prevent creating as many regex instance as you call the method, to compile only once the regex.
- `[\W_]+` transforms `a_b_c` to `Abc`, while the underscore *is* a word separator now, so I would expect `ABC`.
- the possible characters as a regex are `[a-zA-Z_][a-zA-Z_0-9]*`, so `\W` would never be used. (cf [spec](https://avro.apache.org/docs/1.12.0/specification/#names)).

More globally, I think regex here is not a good candidate, as a good old loop over characters could be probably more readable, as many developers doesn't like regex (spoiler: I love that). If the regex is really your go-to, I would mutualize words splitting in a method, so `a_bWord_ZaK` would end up to the list `a`, `b`, `word`, `za`, `k`, and then in pascal case you join and uppercase the first word, camel case lowers the first char, and snake case joins with underscore.

Don't hesitate to make specific unit tests for this class and the different cases. By the way, do not bother testing when special chars are present, as it should not be possible when following avro spec (YAGNI / KISS).

Chuckame (changes requested):

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNamingStrategy.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    public data object CamelCase : FieldNamingStrategy {
+        override fun format(original: String): String = splitWords(original).joinToCamelCase()
+    }
+
+    public data object SnakeCase : FieldNamingStrategy {
+        override fun format(original: String): String = splitWords(original).joinToSnakeCase()
+    }
+
+    public data object PascalCase : FieldNamingStrategy {
```
Chuckame: After thinking, let's remove pascal case and camel case, which are not kotlin's standard. The original need here is that we could have non-standard field names from avro schemas.

gradle-plugin/README.md
```diff
@@ -40,3 +40,44 @@ Then, just use your generated classes in your code (the main sourceSet includes
...
+        fieldNamingStrategy = FieldNamingStrategyType.CAMEL_CASE
+        fieldNamingStrategy = FieldNamingStrategyType.SNAKE_CASE
+        fieldNamingStrategy = FieldNamingStrategyType.PASCAL_CASE
+    }
+}
+```
+
+### Serialization Compatibility
+
+When using a non-identity strategy, `@SerialName` annotations are automatically added to all generated record properties with the original Avro field name. This ensures serialization and deserialization work correctly without any additional configuration.
```
Chuckame: I think it deserves a better explanation on why the annotation is here: to ignore the kotlin's property name and override it with the original field name from the avro schema

Chuckame (changes requested):

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNamingStrategy.kt
Chuckame: Oops, you're right. I was expecting Pascal and snake case to be removed right

gradle-plugin/README.md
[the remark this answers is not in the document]
Chuckame: Sure! Im French, and sometimes I write stuff in English and ask Ai to improve it. Sometimes in French and translate it, sometimes just from a piece of code.

But, the most important part is to not pollute as Ai loves to do: talking too much. Probably you have to ask ai to be concise, or just translate is also enough!

gradle-plugin/README.md
```diff
@@ -40,3 +40,39 @@ Then, just use your generated classes in your code (the main sourceSet includes
...
+    sourcesGeneration {
+        fieldNamingStrategy = FieldNamingStrategyType.IDENTITY    // default
+        fieldNamingStrategy = FieldNamingStrategyType.CAMEL_CASE
+    }
+}
+```
+
+### Serialization Compatibility
+
+Since non-identity strategies alter property names, the generator adds `@SerialName("original_avro_field")` to maintain serialization compatibility. This explicitly maps the Kotlin property back to the original Avro field name, overriding the default naming behavior.
```
Chuckame:
```suggestion
Since changing the naming strategy only changes the kotlin property name but not the avro's record field name, the generator adds `@SerialName("original_avro_field")` to keep the serialized name. This annotation explicitly maps the Kotlin property back to the original Avro field name.
```

kotlin-generator/src/test/kotlin/com/github/avrokotlin/avro4k/FieldNamingStrategyTest.kt
```diff
@@ -0,0 +1,63 @@
+package com.github.avrokotlin.avro4k
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+
+class FieldNamingStrategyTest : StringSpec({
+    "identity keeps input untouched" {
+        listOf(
+            "street_line" to "street_line",
```
Chuckame: same as above here

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNamingStrategy.kt
```diff
@@ -0,0 +1,74 @@
...
+    public data object Identity : FieldNamingStrategy {
+        override fun format(original: String): String = original
+    }
+
+    public data object CamelCase : FieldNamingStrategy {
+        override fun format(original: String): String = splitWords(original).joinToCamelCase()
+    }
+}
+
+private fun splitWords(input: String): List<String> {
```
Chuckame: Now that we have only 1 built-in naming strategy, I propose you to make this method more specific.

As we still have both upper-cammelCase (public) and lower-camelCase (internal), let's at least propose a param to chose if the first letter should be upper or lowercase.

I also feel the current method to be a be complex to read, where I like to create nested functions. Hoping you appreciate it:
```kotlin
public interface FieldNamingStrategy {
    public fun format(original: String): String

    public data object Identity : FieldNamingStrategy {
        override fun format(original: String): String = original
    }

    public data object CamelCase : FieldNamingStrategy {
        override fun format(original: String): String = toCamelCase(original, doUppercaseFirstChar = false)
    }
}

private fun toCamelCase(input: String, doUppercaseFirstChar: Boolean) = StringBuilder(input.length).apply {
    var wordStart = -1

    fun startWord(startInclusive: Int) {
        wordStart = startInclusive
    }

    fun isWordStarted(): Boolean {
        return wordStart != -1
    }

    fun endWord(endExclusive: Int) {
        if (!doUppercaseFirstChar && isEmpty()) {
            append(input.substring(wordStart, endExclusive).lowercase())
        } else {
            append(input[wordStart].uppercaseChar())
            append(input.substring(wordStart + 1, endExclusive).lowercase())
        }
        wordStart = -1
    }

    for (i in input.indices) {
        val c = input[i]
        when {
            !c.isLetterOrDigit() -> {
                if (isWordStarted()) {
                    endWord(i)
                }
            }

            !isWordStarted() -> startWord(i)

            !c.isDigit() && input[i - 1].isDigit() -> {
                // "abc123" -> ["abc", "123"]
                endWord(i)
                startWord(i)
            }

            c.isUpperCase() -> {
                if (!input[i - 1].isUpperCase()) {
                    // "myXml" -> ["my", "Xml"]
                    endWord(i)
                    startWord(i)
                } else if (i + 1 < input.length && input[i + 1].isLowerCase()) {
                    // acronym present: "testXMLParser" -> ["test", "XML", "Parser"]
                    endWord(i)
                    startWord(i)
                }
            }
        }
    }

    if (isWordStarted()) {
        endWord(input.length)
    }
    if (isEmpty()) {
        throw IllegalArgumentException("Input string does not contain any valid letter or digit (a-zA-Z0-9)")
    }
}.toString()

internal fun String.toPascalCase(): String = toCamelCase(this, doUppercaseFirstChar = true)
```

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/KotlinGenerator.kt
```diff
@@ -431,11 +433,13 @@ public class KotlinGenerator(
             .addAnnotationIfNotNull(buildAvroAliasAnnotation(schema))
             .let {
                 schema.fields.fold(it) { builder, field ->
-                    val typeName = getTypeName(field.schema, field.name.uppercaseFirstChar())
+                    val kotlinFieldName = fieldNamingStrategy.format(field.name)
+                    val fieldBaseName = field.name.toPascalCase()
+                    val typeName = getTypeName(field.schema, fieldBaseName)
```
Chuckame: Not sure how helpful is to put it in a separate variable
```suggestion
                    val typeName = getTypeName(field.schema, field.name.toPascalCase())
```

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/KotlinGenerator.kt
```diff
@@ -476,7 +489,8 @@ public class KotlinGenerator(
             .addTypes(
                 schema.fields.mapNotNull { field ->
                     if (field.schema is TypeSafeSchema.UnionSchema) {
-                        generateSealedInterface(field.schema, unionNameFormatter(field.name.uppercaseFirstChar()), field.name.uppercaseFirstChar())
+                        val unionBaseName = field.name.toPascalCase()
+                        generateSealedInterface(field.schema, unionNameFormatter(unionBaseName), unionBaseName)
```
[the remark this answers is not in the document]
Chuckame: For me that is even better following kotlin conventions.

kotlin-generator/src/test/kotlin/com/github/avrokotlin/avro4k/KotlinGeneratorTest.kt
```diff
@@ -78,4 +80,11 @@ class KotlinGeneratorTest {
                 .writeTo(outputBaseDir)
         }
     }
+
+    private fun resolveFieldNamingStrategy(testName: String): FieldNamingStrategy =
+        when (testName) {
+            "field_naming_identity" -> FieldNamingStrategy.Identity
+            "field_naming_camel" -> FieldNamingStrategy.CamelCase
+            else -> FieldNamingStrategy.Identity
+        }
```
Chuckame: As by default the naming strategy is identity, let's only test when camel-case is set (implicitely testing identity for all the other tests):
```suggestion
        when (testName) {
            "field_naming_camel_case" -> FieldNamingStrategy.CamelCase
            else -> FieldNamingStrategy.Identity
        }
```

Chuckame (changes requested):

kotlin-generator/build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
 
     testImplementation(platform(libs.junit.bom))
     testImplementation(libs.kotest.core)
+    testImplementation(libs.kotest.junit5)
```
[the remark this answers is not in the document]
Lee-WonJun: image

Needed for SpringSpec (Kotest) to run on JUnit5. If we’re not using Kotest Spec in this module, I’ll drop it.

gradle-plugin/src/main/kotlin/com/github/avrokotlin/avro4k/plugin/gradle/Avro4kGradlePlugin.kt
```diff
@@ -47,6 +50,12 @@ public interface Avro4kPluginSourcesGenerationExtension {
      */
     public val outputDir: DirectoryProperty
 
+    /**
+     * When true, uses Kotlin naming conventions for generated field names (camelCase).
+     * Defaults to false, keeping original avro field names.
+     */
+    public val useKotlinConventionForFieldNames: Property<Boolean>
```
Chuckame: Just thinking, how would it be rendered in gradle groovy (not kts)? Is it `getUseXx()` or `useXx()`? 🤔

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/FieldNamingStrategy.kt
```diff
@@ -0,0 +1,80 @@
+package com.github.avrokotlin.avro4k
+
+/**
+ * Strategy for converting Avro field names to Kotlin property names.
+ * Only applies to record fields, not to class/enum/union names.
+ */
+public interface FieldNamingStrategy {
```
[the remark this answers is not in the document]
Lee-WonJun: I believe you were referring to `@InternalAvro4kApi`, got it

gradle-plugin/src/main/kotlin/com/github/avrokotlin/avro4k/plugin/gradle/Avro4kGradlePlugin.kt
Lee-WonJun: Interesting. Groovy also supports properties and seems to handle them similarly to Kotlin, so I’d expect we can just use useKotlinConventionForFieldNames in the Groovy DSL as well.

kotlin-generator/src/test/resources/field_naming_camel_case.avsc
```diff
@@ -0,0 +1,51 @@
+{
+    "type": "record",
+    "name": "OriginalCaseRecord",
+    "namespace": "com.github.avrokotlin.avro4k.fieldnaming",
+    "fields": [
+        {
+            "name": "user_id",
```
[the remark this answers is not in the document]
Lee-WonJun: This is a case I somewhat expected to run into.

In other words, I viewed code generation as rule-based: it produces the output first, and then validation is effectively handled by the compiler/build stage.

It’s not exactly the same, but I think it’s similar to how an infix function is treated as a static method internally and ends up with a JVM signature—so if there’s already another static method with the same JVM signature, you get a compilation error.

When I set `useKotlinConventionForFieldNames` to `true`, but if we handle this by falling back and running the identity strategy, that could lead to unexpected behavior. Personally, I don’t think that’s desirable.

So if we decide to handle this issue, I think it would be better to detect the collision explicitly and fail fast with a clear error.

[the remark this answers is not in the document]
Chuckame: Oh wait. It actually (should) fail, hopefully already done by kotlinpoet validation.

The point here is: do we want to fail or not? If yes, it's actually managed, so the user will have to keep the default naming strategy to not fail (just to check if the exception is clear enough).

If no failure expected, like a lenient mode, we need to switch the naming strategy to identity on the fly. But we also need to decide if we want to fallback on identity for the file, or for just the corresponding fields (I would prefer the first).

Now, with all of that in mind, putting back on the table the YAGNI-way: do we want to handle it now? As it seems quite a rare issue. However, it would be interesting to check the exception if it's clear enough. What do you think?

kotlin-generator/src/test/resources/field-naming-camel-case.avsc
```diff
@@ -0,0 +1,51 @@
+{
+    "type": "record",
+    "name": "OriginalCaseRecord",
+    "namespace": "com.github.avrokotlin.avro4k.fieldnaming",
+    "fields": [
+        {
+            "name": "user_id",
```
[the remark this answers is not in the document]
Lee-WonJun: 1. **At the moment, it seems that even without any special handling, this does not get caught by `kotlinpoet` validation.**
   I ran a quick temporary test to check what error message we would get, but the file is generated anyway, and `@SerialName` also seems to get partially messed up. In other words, generation itself still succeeds.

2. **If we decide to fail, we should surface the duplicated/conflicting fields clearly.**
   When naming issues cause problems in the consumer project, users are likely aware they enabled this flag, so they will probably fix the schema or adjust their conventions on their side.
   If we raise an error, something along the lines of `Conflicting declarations: ...` would be a good message, as it directly points to the conflict.

3. **If we support a lenient mode, then on failure we should fallback at the *avsc file level*.**
   A field-level fallback would make the rules far too difficult to understand when looking at the generated code (the problem is not implementation complexity, but comprehension complexity).
   Also, because lenient mode would not raise an error, users might interpret “why only this field didn’t change?” as a bug and open issues.

4. **I consider this a rare issue.**
   Even if we do nothing, it is often caught in the user’s build/compile phase (and in many cases, as soon as the class is used in the IDE).
   Additionally, the camelCase option is non-default (i.e., an advanced feature), so I do not expect this to occur frequently.

4-2. **I also don’t think fallback should be the default behavior.**
If we make it configurable via another Gradle option, that’s possible, but it increases configuration complexity.
For example, if a user accidentally removes the field that previously triggered fallback, the generated Kotlin class would start using camelCase again. That’s confusing.
Given this is codegen, it seems better to avoid fallback and fail explicitly instead.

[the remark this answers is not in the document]
Chuckame: Good points, especially about when the user removes a field, so the generation gets back to camel case.

So very simple solution: when adding a field to a record, we could simply check if it already exists. Check the kdoc part: when I add kdoc from avro default value, I check if it already has kdoc, so I add proper line break.

Maybe we can do that (I'll check)

kotlin-generator/src/main/kotlin/com/github/avrokotlin/avro4k/KotlinGenerator.kt
```diff
@@ -445,11 +447,12 @@ public class KotlinGenerator(
...
                 schema.fields.fold(it) { builder, field ->
-                    val typeName = getTypeName(field.schema, field.name.uppercaseFirstChar())
+                    val kotlinFieldName = fieldNamingStrategy.format(field.name)
+                    val typeName = getTypeName(field.schema, field.name.toPascalCase())
 
                     builder.addPrimaryProperty(
-                        PropertySpec.builder(field.name, typeName.typeName)
-                            .initializer(field.name)
+                        PropertySpec.builder(kotlinFieldName, typeName.typeName)
+                            .initializer(kotlinFieldName)
```
Chuckame: Here it is:
```suggestion
                    val kotlinFieldName = fieldNamingStrategy.format(field.name)
                    val typeName = getTypeName(field.schema, field.name.toPascalCase())

                    require(it.propertySpecs.none { it.name == kotlinFieldName }) {
                        "The record ${schema.fullName} contains duplicated fields when applying custom naming strategy. " +
                                "The actual avro field ${field.name} has been mapped to $kotlinFieldName which has already been added. " +
                                "Schema: $schema"
                    }

builder.addPrimaryProperty(
                        PropertySpec.builder(kotlinFieldName, typeName.typeName)
                            .initializer(kotlinFieldName)
```

kotlin-generator/src/test/resources/field-naming-camel-case.avsc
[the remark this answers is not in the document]
Lee-WonJun: I committed your suggestion and now see an error for the `field-naming-camel-case-fail.avsc` test case.

The message is detailed and helpful. It would be nicer if it showed which original field produced the conflicting name, but that feels too costly to implement for the benefit.
```
The record ns.OriginalCaseRecord contains duplicated fields when applying custom naming strategy. The actual avro field userId has been mapped to userId which has already been added. Schema: RecordSchema(originalSchema={"type":"record","name":"OriginalCaseRecord","namespace":"ns","fields":[{"name":"user_id","type":"int","doc":"User identifier","default":0,"aliases":["user_identifier"]},{"name":"userId","type":"int","doc":"User identifier - camelCased","default":0,"aliases":["user_identifier_camelCased"]},{"name":"billing_address","type":{"type":"record","name":"BillingAddress","fields":[{"name":"street_line","type":"string","doc":"Street line","default":"unknown"}]},"doc":"Full billing address"},{"name":"account_status","type":["string",{"type":"enum","name":"StatusFlag","symbols":["ACTIVE","INACTIVE"]},"null"],"doc":"Status wrapper","aliases":["status_alias"]}]}, name=OriginalCaseRecord, space=ns, fields=[Field(name=user_id, schema=IntSchema(originalSchema="int", isNullable=false, props={}), defaultValue=0, doc=User identifier, aliases=[user_identifier], props={}), Field(name=userId, schema=IntSchema(originalSchema="int", isNullable=false, props={}), defaultValue=0, doc=User identifier - camelCased, aliases=[user_identifier_camelCased], props={}), Field(name=billing_address, schema=RecordSchema(originalSchema={"type":"record","name":"BillingAddress","namespace":"ns","fields":[{"name":"street_line","type":"string","doc":"Street line","default":"unknown"}]}, name=BillingAddress, space=ns, fields=[Field(name=street_line, schema=StringSchema(originalSchema="string", isNullable=false, props={}), defaultValue=unknown, doc=Street line, aliases=[], props={})], doc=null, aliases=[], isNullable=false, props={}), defaultValue=java.lang.Object@44841b43, doc=Full billing address, aliases=[], props={}), Field(name=account_status, schema=UnionSchema(originalSchema=["string",{"type":"enum","name":"StatusFlag","namespace":"ns","symbols":["ACTIVE","INACTIVE"]}], types=[StringSchema(originalSchema="string", isNullable=false, props={}), EnumSchema(originalSchema={"type":"enum","name":"StatusFlag","namespace":"ns","symbols":["ACTIVE","INACTIVE"]}, name=StatusFlag, space=ns, symbols=[ACTIVE, INACTIVE], defaultSymbol=null, doc=null, aliases=[], isNullable=false, props={})], isNullable=true, props={}), defaultValue=java.lang.Object@44841b43, doc=Status wrapper, aliases=[status_alias], props={})], doc=null, aliases=[], isNullable=false, props={})
```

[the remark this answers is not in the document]
Chuckame: Agree it's not the best error explanation, but as it's probably a rare use case, let's finish here, and iterate over if not enough from users 👌

As far as I'm thinking, you could gather avro field names corresponding to the failing kotlin name with `schema.fields.map { it.name to fieldNamingStrategy.format(it.name) }.groubBy { it.value }[kotlinFieldName]`, and pass it to the error message to say "hey, here are the conflicting original avro names"

gradle-plugin/README.md
```diff
@@ -213,3 +213,29 @@ Here is an example of the schema definition:
     "symbols": ["FR", "GB", "IT"]
 }
 ```
+
+## Kotlin Field Name Convention
```
Chuckame:
```suggestion
### Comply to Kotlin property naming convention
```

Chuckame (changes requested):
I did the last pass, for improving a bit the docs after the rebase. Then we can merge !!!

Chuckame (approved):
Many thanks for the PR! Don't hesitate if you want to contribute more on other aspects!

GuillaumeGuerra:
hi everyone
do you have a rough estimate of when would this feature be delivered in an upcoming release ?
thanks a lot

Lee-WonJun:
> hi everyone do you have a rough estimate of when would this feature be delivered in an upcoming release ? thanks a lot

Isn’t this already available?

Chuckame:
@GuillaumeGuerra see https://github.com/avro-kotlin/avro4k/issues/425#issuecomment-3786884084

GuillaumeGuerra:
you're absolutely right
somehow I messed up with the released version in my gradle script file

thanks again for the feature, it's very helpful 🙏

GuillaumeGuerra:
actually, I understand why I messed up !

I blindly copy/pasted the gradle-plugin sample available [here](https://github.com/avro-kotlin/avro4k/tree/main/gradle-plugin#setup), which still points at version 2.5.3
I did not verify there was no new version, I assumed that version was compatible with the new feature ^^

PR #465, opened 2026-04-06
fix(kotlin-generator): prevent stackoverflow where a record is used inside itself in a simple/nullable union

sanderv:
This would generate a StackOverflowException in the previous code.


Chuckame:
Please run `actionsBeforeCommit` to fix the build error (I'll add it in a git hook soon + CONTRIBUTOR.md guide!)

PR #478, opened 2026-04-26
feat(core): support kotlin.time.Instant

### LINKED ISSUE #417: Provide kotlin.time.Instant serializer
Today, only java's time module in handled by default in avro4k.

Make it handling `kotlin.time.Instant` module, which is kotlin native, and fits the future goal of making avro4k truly multiplatform. The serializer is barely 95% the same as `UUIDSerializer` for the java's UUID type.

It also has to match the logical types `timestamp-millis`, `timestamp-micros` and `timestamp-nanos`.

Do not forget to register it in the `Avro` default serializers module.

Special mention: it already have a `Uuid.serializer()`, but we have to override it as it only manages a String value, and is not capable of generating a schema. Look at `SerializerLocatorMiddleware` on how to override it.

kangwooc:
## Summary
- Add native serializer support for `kotlin.time.Instant` without adding a `kotlinx-datetime` dependency.
- Encode as `timestamp-millis` by default, with writer-schema support for `timestamp-micros`, `timestamp-nanos`, and ISO-8601 strings.
- Update the public API dump and reject out-of-range Avro timestamp conversions instead of overflowing or clamping.

## Testing
- `./gradlew :core:test --tests '*KotlinInstantEncodingTest' --tests '*KotlinInstantDecodingTest' :core:apiCheck`
- `./gradlew check`

Closes #417


Chuckame:
I've been near to merge your PR, however I would have expect the std `kotlin.time.Instant` to be used, instead of the "additional" other Instant from the kotlinx datetime library, to not depend on another library (not all kotlin projects needs the kotlinX datetime library).

Please adapt it to this `kotlin.time.Instant`.

If we really want to add support for kotlinx datetime, let's try to propose a modular approach of serializing a timestamp millis/micros/nanos, which would be used by jvm's Instant/Date, the kotlin's Instant, or any other non-kotlin standard time modules. It could be then used on the user-application's side (preferred solution for uncommon time libraries), another avro4k module (preferred for common/official kotlin's libraries, like ktor does), or included in the core lib but optionally (less preferred). Let's put this modularity into another PR 🙏 

Thanks again for your work!

kangwooc:
@Chuckame then do i create another pr and closed this?  
if you are okay, i could keep working on this pr.

Chuckame:
As you prefer, you can reuse this PR also, as mostly only the package name will change in the changes.

README.md
```diff
@@ -372,7 +372,7 @@ Add the avro4k dependency:
 
 | Avro4k     | Kotlin   | Kotlin API/language | Kotlin serialization |
 |------------|----------|---------------------|----------------------|
-| `>= 2.0.0` | `>= 2.0` | `>= 1.9`            | `>= 1.7.0`           |
+| `>= 2.0.0` | `>= 2.3` | `>= 1.9`            | `>= 1.7.0`           |
```
kangwooc: I changed this because `kotlin.time.Instant` is stable in the Kotlin stdlib from Kotlin 2.3, and this PR exposes it in the public API.

  But I agree the compatibility table should probably stay as the general Avro4k requirement. I’ll keep it as `>= 2.0` and document Kotlin `>= 2.3` only as a requirement for native
  `kotlin.time.Instant` support.

Chuckame: Good catch. 

If the user wants to use this instant, then he needs to be in 2.3, but I think that it should still work with 2.0, just that it would fail if the user tried to directly use the instant serializer class. In spring, they do it massively, so there are many classes based on a class that may not be in the classpath for auto configuration. It would fail only when using directly the class.

You could try to reference from another project your avro4k build but using kotlin 2.0 to see if it works.

Regarding when it became stable, I think it does not matter: we are only providing a serializer, so none of the codebase actually relies on this class. Maybe the middleware could fail here, let's double check that with an earlier version of kotlin if it still works

kangwooc: Thanks, I tested this locally.

With a Java runtime consumer and `kotlin-stdlib` forced to `2.0.21`, just loading `Avro.Default` fails with `NoClassDefFoundError: kotlin/time/Instant`. So in the current implementation, it does not only fail when directly using the Instant serializer; `KotlinStdLibSerializersModule` registers `KotlinInstantSerializer` eagerly, and the middleware directly references `Instant.serializer()`.

So I think we either need to keep Kotlin `>= 2.3`, or make the Instant support lazy/guarded so older stdlib versions can still load Avro4k and only fail when `kotlin.time.Instant` is actually used.

core/src/main/kotlin/com/github/avrokotlin/avro4k/serializer/KotlinStdLibrarySerializers.kt
```diff
@@ -20,12 +21,14 @@ import kotlinx.serialization.modules.SerializersModule
...
 import org.apache.avro.LogicalType
 import org.apache.avro.Schema
+import kotlin.time.Instant
 import kotlin.uuid.Uuid
 
 internal val KotlinStdLibSerializersModule: SerializersModule
     get() =
         SerializersModule {
             contextual(KotlinUuidSerializer)
+            contextual(KotlinInstantSerializer)
```
Chuckame: Here you could put a runCatching to allow not having `Instant` in the classpath

core/src/main/kotlin/com/github/avrokotlin/avro4k/internal/SerializerLocatorMiddleware.kt
```diff
@@ -44,6 +46,7 @@ internal object SerializerLocatorMiddleware {
             serializer === ByteArraySerializer() -> AvroByteArraySerializer
             serializer === Duration.serializer() -> KotlinDurationSerializer
             serializer === Uuid.serializer() -> KotlinUuidSerializer
+            serializer === Instant.serializer() -> KotlinInstantSerializer
             else -> serializer
```
Chuckame: Here is the second issue that would appear after the serializersModule is fixed.

But also, having this `when` growing, it creates many branches, which is not ideal for keeping good performances, I'm thinking that this could be put inside an `IdentityHashMap`. This would enable linear lookup. To be tested with the current benchmarks what would be the impact.

So a map could be built once, including — or not, classes that may be absent from the classpath.

README.md
Chuckame: I did comment the 2 places where it should fail 👀

Chuckame (changes requested):

README.md
```diff
@@ -375,6 +375,10 @@ Add the avro4k dependency:
 | `>= 2.0.0` | `>= 2.0` | `>= 1.9`            | `>= 1.7.0`           |
 | `< 2.0.0`  | `>= 1.6` | `>= 1.6`            | `>= 1.3`             |
 
+> [!NOTE]
```
kangwooc, line 378: I added the comment for compatibility with using `kotlin.time.Instant`

kangwooc:
@Chuckame just for reminder, would you review this?

core/src/test/kotlin/com/github/avrokotlin/avro4k/OptionalKotlinInstantClasspathTest.kt
```diff
@@ -0,0 +1,57 @@
+package com.github.avrokotlin.avro4k
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+import java.io.File
+import java.net.URL
+import java.net.URLClassLoader
+
+internal class OptionalKotlinInstantClasspathTest : StringSpec({
+    "load Avro and generate schemas when kotlin.time.Instant is absent but unused" {
```
Chuckame: Would be great to also add encoding/decoding tests, as Instant may be loaded in other places. Just encode decode an Int

core/src/test/kotlin/com/github/avrokotlin/avro4k/OptionalKotlinInstantClasspathTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return System.getProperty("java.class.path")
+        .split(File.pathSeparator)
+        .map(::File)
+        .filter(File::exists)
+        .map { it.toURI().toURL() }
+        .toTypedArray()
+}
+
+private class OptionalKotlinInstantBlockingClassLoader(urls: Array<URL>) :
```
Chuckame: Let's rename it to clearer `ClassLoaderWithoutKotlinInstant`

core/src/test/kotlin/com/github/avrokotlin/avro4k/OptionalKotlinInstantClasspathTest.kt
Chuckame: Also rephrase this test, as yes it loads Avro without kotlin instant, but what is expected and why this test exists?

Probably something like `Loading Avro instance and generating a schema should not fail if kotlin's Instant is missing from the class path`

core/src/main/kotlin/com/github/avrokotlin/avro4k/internal/SerializerLocatorMiddleware.kt
```diff
@@ -39,34 +42,44 @@ import kotlin.uuid.Uuid
...
+        this[Uuid.serializer()] = KotlinUuidSerializer
+        runCatching { this[Instant.serializer()] = KotlinInstantSerializer }
+    }
+
+    private val deserializers = IdentityHashMap<DeserializationStrategy<*>, DeserializationStrategy<*>>().apply {
+        this[ByteArraySerializer()] = AvroByteArraySerializer
+        this[Duration.serializer()] = KotlinDurationSerializer
+        this[Uuid.serializer()] = KotlinUuidSerializer
+        runCatching { this[Instant.serializer()] = KotlinInstantSerializer }
+    }
```
Chuckame: You can use only one map, as they are always casted after. Don't forget the *in* keyword to allow sub types of KSerializer for the lookup.
```suggestion
    private val serializers: Map<in KSerializer<*>, KSerializer<*>> = IdentityHashMap().apply {
        this[ByteArraySerializer()] = AvroByteArraySerializer
        this[Duration.serializer()] = KotlinDurationSerializer
        this[Uuid.serializer()] = KotlinUuidSerializer
        runCatching { this[Instant.serializer()] = KotlinInstantSerializer }
    }
```

then
```
    fun <T> apply(serializer: SerializationStrategy<T>): SerializationStrategy<T> {
        serializers[serializer]?.let { return it as SerializationStrategy<T> }

        return serializer
    }

    @OptIn(InternalSerializationApi::class)
    fun <T> apply(deserializer: DeserializationStrategy<T>): DeserializationStrategy<T> {
        serializers[deserializer]?.let { return it as DeserializationStrategy<T> }
        (deserializer as? AbstractCollectionSerializer<*, T, *>)?.let { return AvroCollectionSerializer(deserializer) }

        return deserializer
    }
```

core/src/main/kotlin/com/github/avrokotlin/avro4k/serializer/KotlinStdLibrarySerializers.kt
```diff
@@ -108,4 +111,139 @@ public object KotlinUuidSerializer : AvroSerializer<Uuid>(Uuid::class.qualifiedN
...
+        } catch (e: ArithmeticException) {
+            throw SerializationException("$this is out of range for Avro $logicalType", e)
+        }
+    }
+
+    private fun Instant.Companion.fromEpochNanos(nanos: Long): Instant {
+        val seconds = nanos / NANOS_PER_SECOND
+        val nanosOfSecond = nanos % NANOS_PER_SECOND
+        return fromEpochSeconds(seconds, nanosOfSecond)
+    }
```
Chuckame: I propose you to regroup companion's extensions `Instant.fromXxx`.

You should also consider negative instants, which ended up to wrong tests.

Maybe the name toAvroTimestamp could be renamed to toEpochUnits or similar.
```
    private fun Instant.toEpochUnits(
        unitsPerSecond: Long,
        logicalType: String,
    ): Long {
        try {
          val nanosPerUnit = NANOS_PER_SECOND / unitsPerSecond // the JIT should do its work to inline the computed value, as unitsPerSecond has only 3 possible values
          if (epochSeconds < 0 && nanosecondsOfSecond > 0) {
              val units = Math.multiplyExact(epochSeconds + 1, unitsPerSecond)
              val adjustment = nanosecondsOfSecond / nanosPerUnit - unitsPerSecond
              return Math.addExact(units, adjustment)
          } else {
              val units = Math.multiplyExact(epochSeconds, unitsPerSecond)
              return Math.addExact(units, nanosecondsOfSecond / nanosPerUnit)
          }
        } catch (e: ArithmeticException) {
            throw SerializationException("$this is out of range for Avro $logicalType", e)
        }
    }
```

I would put all those extensions in another file, with `internal` keyword, to be more extensively tested separately to not pollute too much the encoding/decoding avro tests.

Here the code from the java's std `toEpochMillis` as a reference:
```
        if (seconds < 0 && nanos > 0) {
            long millis = Math.multiplyExact(seconds+1, 1000);
            long adjustment = nanos / 1000_000 - 1000;
            return Math.addExact(millis, adjustment);
        } else {
            long millis = Math.multiplyExact(seconds, 1000);
            return Math.addExact(millis, nanos / 1000_000);
        }
```

Chuckame (changes requested):
Near to final code, sorry for the delay again

core/src/main/kotlin/com/github/avrokotlin/avro4k/serializer/KotlinInstantExtensions.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.serialization.SerializationException
+import kotlin.time.Instant
+
+internal const val NANOSECONDS_PER_MILLISECOND: Long = 1_000_000L
+internal const val NANOSECONDS_PER_MICROSECOND: Long = 1_000L
+internal const val MILLIS_PER_SECOND: Long = 1_000L
+internal const val MICROS_PER_SECOND: Long = 1_000_000L
+internal const val NANOSECONDS_PER_SECOND: Long = 1_000_000_000L
+
+internal fun Instant.toEpochUnits(
```
Chuckame: Would be better to keep as you did with one central private function, and 3 extensions function (toEpochXx, even naming them toExactEpochXx to not be confused with the stdlib toEpochMilliseconds), so this would make all the NANOSECONDS_PER_MILLISECOND and similar constants private to not leak them out of this file

kangwooc: got it. I changed it

Chuckame (approved):

PR #485, opened 2026-06-14
feat(core): enable zero-copy decoding for jvm ByteBuffer

### LINKED ISSUE #476: Support for zero-copy deserialization

Kafka consumers read messages in batches, under the hood kafka-clients library allocates a single `ByteBuffer` for the data received from the broker and then maps bytes to internal data structures and records (zero-copy). When it comes to deserializing an individual record, the library passes a `ByteBuffer` slice to the deserializer implementation exposing only the bytes of that individual record. Most of the deserializers just copy it to a byte array and do their work (see `default T deserialize(String topic, Headers headers, ByteBuffer data)` method in `org.apache.kafka.common.serialization.Deserializer`) 

I was wondering if it's possible to support zero-copy deserialization from `ByteBuffers` to Avro objects directly. The first 4 bytes contain schemaId (added by KafkaAvroSerializer) which is retrieved from a cache or Schema Registry, so we can assume we have an Avro schema and a `ByteBuffer` with the Avro data.

Zero-copy deserialization for Avro data. 

**Describe alternatives you've considered**
–

**Additional context**
–

Chuckame:
Close #476

### DIFF
--- a/core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroJVMExtensions.kt
+++ b/core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroJVMExtensions.kt
@@ -1,13 +1,18 @@
 package com.github.avrokotlin.avro4k
 
+import com.github.avrokotlin.avro4k.internal.decodeWithApacheDecoder
 import kotlinx.io.asSink
 import kotlinx.io.asSource
 import kotlinx.io.buffered
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.serializer
 import org.apache.avro.Schema
+import org.apache.avro.io.DecoderFactory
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroJVMExtensions.kt
```diff
@@ -62,4 +67,26 @@ public inline fun <reified T> Avro.decodeFromStream(
     inputStream: InputStream,
 ): T {
     return decodeFromSource(writerSchema, inputStream.asSource().buffered())
+}
+
+/**
+ * Decode the value [T] from the given ByteBuffer.
```
Chuckame:
```suggestion
 * Decode the value [T] from the given ByteBuffer, helpful when expecting zero-copy decoding from the jvm.
```

PR #505, opened 2026-09-05
feat(core): Allow specifying sync marker for byte deterministic output

### LINKED ISSUE #83: encodeToByteArray() non deterministic?
While ser/de works, the ByteArray returned by `encodeToByteArray` seems no deterministic.

For example:
```
@Serializable
data class TaskOptions(
    val runningTimeout: Float? = null
)

fun main() {
    val m = TaskOptions(2F)
    val b1 = Avro.default.encodeToByteArray(TaskOptions.serializer(), m)
    val b2 = Avro.default.encodeToByteArray(TaskOptions.serializer(), m)
    println(b1.contentEquals(b2))
}
```
will print "false". 

Besides being strange, it's an issue for some unit tests

mdedetrich:
Resolves https://github.com/avro-kotlin/avro4k/issues/83, see justification at https://github.com/avro-kotlin/avro4k/issues/83#issuecomment-5554479089.

This PR allows providing a sync marker in order to provide byte deterministic output of avro files.

A test has been written to confirm that everything works as expected.

### DIFF
--- a/core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroObjectContainer.kt
+++ b/core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroObjectContainer.kt
@@ -7,6 +7,7 @@ import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.serializer
 import org.apache.avro.Schema
 import org.apache.avro.file.CodecFactory
+import org.apache.avro.file.DataFileConstants
 import org.apache.avro.file.DataFileStream
 import org.apache.avro.file.DataFileWriter
 import org.apache.avro.io.DatumReader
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/github/avrokotlin/avro4k/AvroObjectContainer.kt
```diff
@@ -102,6 +109,20 @@ public inline fun <reified T> AvroObjectContainer.decodeFromStream(
...
 @ExperimentalAvro4kApi
 public class AvroObjectContainerBuilder internal constructor(private val fileWriter: DataFileWriter<*>) {
+    internal var syncMarker: ByteArray? = null
+        private set
+
+    /**
+     * Fixes the 16-byte block sync marker instead of generating a random one, making the output byte-for-byte deterministic.
+     * Prefer a random-looking value so readers can still resynchronize within a stream.
+     */
+    public fun syncMarker(value: ByteArray) {
```
Chuckame: Could you add a link from the avro spec (if any) to explain a bit more what's for?

Chuckame: I also might prefer the explanation done in the readme that you could reuse here:

```
By default a random 16-byte sync marker is generated per file, so encoding the same data twice produces different bytes.
Setting a `syncMarker` replace the use of random sync markers by the given static ByteArray to get byte-for-byte deterministic output (useful for tests or content-addressed storage)
```

mdedetrich: I have just added a commit that does a better job of clarifying the `syncMarker`. Let me know if any further improvements are needed

Chuckame (approved):
Let's merge!

